PR #66, opened 2026-07-27
Document PyPI trusted publisher setup

md-mt:
[Assisted by Codex]

## Summary
- document the exact PyPI trusted publisher values required by the release workflow
- add a regression test so the release docs keep the owner/repo/workflow/environment values visible

## Context
The v0.2.0 release workflow built artifacts and created the GitHub release, but PyPI upload failed with invalid-publisher because PyPI has no matching trusted publisher for repo md-mt/termproof, workflow release.yml, environment pypi.

## Verification
- uv run python -m unittest tests.test_release_docs
- uv run python -m unittest discover -s tests

Refs #7

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -34,6 +34,20 @@ uv run termproof run examples/generic examples/multi_turn_conversation.recipe.js
    and release body, uploads evidence, and creates a GitHub release.
 4. PyPI publishing uses GitHub trusted publishing from the release workflow.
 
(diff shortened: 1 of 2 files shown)


md-mt (commented):
[Assisted by Codex] Reviewed the PyPI trusted-publishing docs and regression test. Local full tests passed and GitHub checks are green. No blocking issues found for this docs change.

PR #67, opened 2026-07-27
Make Pages deploy opt-in

md-mt:
[Assisted by Codex]

## Summary
- require repository variable ENABLE_PAGES=true before running actions/deploy-pages
- keep the Pages build and preview artifact running by default
- add regression coverage so public repository visibility does not bypass the opt-in guard

## Context
The latest main push failed in the Pages workflow because actions/deploy-pages returned 404 when GitHub Pages was not enabled. The previous guard deployed automatically for public repos, but Pages still has to be enabled in repository settings.

## Verification
- uv run python -m unittest tests.test_docs_pages
- uv run python -m unittest discover -s tests

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -111,23 +111,17 @@ jobs:
           path: _site
           if-no-files-found: warn
 
-  # Deployment is guarded so private repos without Pages enabled do not fail CI.
-  # For private repos, set repository variable ENABLE_PAGES=true to enable,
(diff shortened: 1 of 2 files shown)


md-mt (commented):
[Assisted by Codex] Reviewed the Pages workflow guard and regression tests. Local full suite passed; GitHub checks are green, with Pages deploy skipped as intended when ENABLE_PAGES is not true. No blocking issues found.

PR #149, opened 2026-08-07
fix(session): arm wait_for_idle on the session's first byte of output

md-mt:
## Intent

Act on the human code review of PR 149 (branch fm/flaky-pty-p4, review report at /Users/mengwei/firstmate/data/review-149-r1/report.md, posted as a PR comment). The PR's core fix is confirmed correct and is NOT being redone: wait_for_idle() must not arm its stable window before the child's first byte, otherwise a slow-starting target has its blank startup screen recorded as the final evidence. The reviewer independently reproduced the product bug through the real CLI on main (0/1 passed, empty final.txt, 124-byte header-only cast) and confirmed the regression tests genuinely fail without the production change.

This round corrects one overreach and documents a deliberate trade:

1. (major) Narrow the idle detector. The previous version used raw byte length both to ARM the stable window and as an ONGOING change detector. Arming with it is deliberate and correct - a first byte that is pure escape sequence never changes screen_text, so the timer would never start. Using it as an ongoing detector overreached: any TUI emitting bytes without changing rendered characters (terminal-title ticks, colour-only animation, idempotent repaints) could never go idle and burned the full timeout. The condition is now 'current != last_screen or (stable_since is None and raw_len != last_raw_len)', exactly the narrowing the reviewer verified. A new QuiescenceBehaviorTest case pins it with an OSC-title-ticking child; it fails (False after the full 3s) without the narrowing.

2. (major) Document the silent-target trade rather than silence it. A target that stays alive and emits nothing can no longer report idle, and 'termproof init' scaffolds wait_for_idle as the only step, so a new user hits this first. The reviewer's judgement, accepted deliberately: for an evidence-recording tool a zero-output session is not something we can attest to, so it is reported rather than passed. This is INTENDED behaviour, not a bug to fix. It is now made explicit via a CHANGELOG '### Changed' entry, a README note in the config block, and a failure detail that distinguishes the case - 'no output observed from the session' when nothing was ever received, versus the existing 'timed out waiting for idle' when output arrived but never settled.

3. (minor) Document, not fix, the uncapped-timeout cost: with idle_cap_seconds null and a silent target the trailing quiescence wait now costs the full recipe timeout. That is literally consistent with the documented semantics of null, so it is one README sentence, deliberately not a code change.

4. (minor) Correct AGENTS.md. The earlier draft blamed pytest's harness startup for shifting PTY timing and told contributors to run PTY tests under both runners. The reviewer disproved the mechanism: the trigger was 'uv run --with pytest' building an ephemeral overlay interpreter per invocation (0.347s to first PTY byte cold vs 0.118s warm, against stable_seconds 0.3); installing pytest into the project venv makes the identical command pass, and the cited example also fails under plain unittest. pytest is deliberately NOT being added to the project - it is not in the dev group and CI never runs it, so the doc would prescribe an unenforceable gate. AGENTS.md now keeps only the durable lesson: a PTY test must never depend on first output arriving within stable_seconds.

5. (nit) Tidy First commit hygiene, explicitly requested: the branch was rebased onto origin/main and its history restructured so the two-line behavioural change, the step failure-message change, the docs, the .gitignore entry, and AGENTS.md are five separate commits, readable and revertable on their own. All content from the prior pipeline fix commits is preserved in the new commits (the seed-from-prior-activity fix and its already_quiet regression test are in the session.py commit); only the AGENTS.md text was intentionally rewritten per finding 4. Same branch, same PR - no new PR, no merge.

Deliberately out of scope, per the review: the residual premature-idle race for a target with a quiet gap mid-startup (inherent to timer-based idle detection, explicitly not this PR's job), and the identical bug in the Rust port (tracked in the RUST-0xx parity workstream). No existing regression test was weakened or deleted - both directions of the new behaviour stay pinned.

Verification reproduced locally: all 5 QuiescenceBehaviorTest tests pass; behaviour probes through real TerminalSession with wait_for_idle(0.75, 3.0) give OSC-title True@0.89s, colour-only True@0.86s, idempotent-redraw True@0.87s, silent-alive False@3.02s (correct), all with the child verified still alive - and without the narrowing the first three regress to False@~3.0s. The slow-start scenario end to end through the CLI: with the fix 1/1 passed, final.txt 'SERVICE READY on port 8080', 176-byte cast; with origin/main's session.py 0/1 passed, empty final.txt, 124-byte cast. Silent target through the CLI now reports 'no output observed from the session'. Full suite: 388 tests OK; ruff and mypy clean.

## What Changed

- `TerminalSession.wait_for_idle()` no longer starts its stable window before the session has produced anything: `stable_since` begins as `None` unless prior output already exists, and is armed by a change in raw output length. A slow-starting target can no longer have its blank startup screen recorded as idle. Raw length only *arms* the window — the ongoing change detector remains rendered screen text (`current != last_screen or (stable_since is None and raw_len != last_raw_len)`), so terminal-title ticks, colour-only animation and idempotent repaints still go idle instead of burning the full timeout.
- The `wait_for_idle` step now reports `no output observed from the session` when the session emitted nothing at all, distinguishing a silent target from the existing `timed out waiting for idle` when output arrived but never settled.
- Docs record the deliberate trade — a CHANGELOG `### Changed` entry, a README note in the config block (including that a silent target with `idle_cap_seconds: null` burns the full recipe timeout), and a corrected `wait_for_idle` description on the Pages site. New `QuiescenceBehaviorTest` cases pin both directions (pre-first-output, OSC-only arming, title-ticking child, already-quiet session), `tests/test_wait_for_idle_step.py` pins the three failure/success details, `AGENTS.md` records the PTY-test invariant, and `.gitignore` excludes `.audit/`.

## Risk Assessment

✅ Low: The behavioural surface is two lines in wait_for_idle plus one failure-detail branch; both directions of the new semantics are pinned by discriminating regression tests, the only other evidence-capture paths gate on wait_for_exit and are unaffected, and the docs now match the implemented behaviour, leaving only a test-timing nit and an out-of-scope Rust parity note.

## Testing

Ran the targeted quiescence and step-detail unit tests plus the session/runner and docs/scaffold modules (all pass, and the quiescence tests were stable over 5 repeats), after fixing a PATH-only setup issue where the PTY tests need the venv's asciinema CLI. Beyond unit tests I reproduced the product behavior through the real CLI: a slow-starting daemon records an empty final.txt and a 123-byte header-only cast on origin/main's session.py but records 'SERVICE READY on port 8080' and passes 1/1 with this PR (captured as before/after PNG screenshots of TermProof's own final-evidence surface); an OSC-title-ticking TUI reaches idle with the narrowed condition and times out without it; the untouched `termproof init` scaffold against a silent target now fails with 'no output observed from the session'; and the documented uncapped-idle cost measured 13.2s with `idle_cap_seconds: null` versus 4.1s at 3.0. Each new regression test was also confirmed to fail against the specific pre-fix variant it guards, and the worktree was left clean.

- Evidence: Final recorded evidence — BEFORE (origin/main session.py): blank screen (local file: <code>/var/folders/00/dn1zh_wx3ys5gq4f1dtgpp8h0000gn/T/no-mistakes-evidence/01KZM0D111VMYBP3B7ZV2BMBV8/final-evidence-BEFORE-origin-main.png</code>)
- Evidence: Final recorded evidence — AFTER (this PR): readiness banner captured (local file: <code>/var/folders/00/dn1zh_wx3ys5gq4f1dtgpp8h0000gn/T/no-mistakes-evidence/01KZM0D111VMYBP3B7ZV2BMBV8/final-evidence-AFTER-this-pr.png</code>)
<details>
<summary>Evidence: Full end-to-end CLI evidence transcript (all four scenarios, before/after)</summary>

```text
# TermProof PR 149 — end-to-end CLI evidence

Recipes live in `recipes/`. All runs are real `termproof run` invocations against
PTY targets; only `termproof/session.py` was swapped between the "before" and
"after" runs (restored afterwards, worktree clean).

## 1. Slow-starting daemon — the product bug the PR fixes

Target: prints nothing for 2s, then `SERVICE READY on port 8080`, then keeps serving.
Recipe: single `wait_for_idle` step (`stable_seconds: 0.5`), `expect_exit_code: null`.

### BEFORE — `termproof/session.py` from origin/main (080a3b2)

`` `
0/1 passed
FAIL slow-start-daemon [default]

### Assertions
- FAIL `output_contains` - contains 'SERVICE READY on port 8080'

### Steps
- PASS `1:wait_for_idle` - stable for 0.5s      <-- armed on the blank startup screen

final.txt      = 1 byte (empty)
session.cast   = 123 bytes (header only, nothing recorded)
`` `

### AFTER — this PR

`` `
1/1 passed
PASS slow-start-daemon [default]

### Assertions
- PASS `output_contains` - contains 'SERVICE READY on port 8080'

### Steps
- PASS `1:wait_for_idle` - stable for 0.5s      <-- armed at the first byte

final.txt      = 27 bytes -> "SERVICE READY on port 8080"
session.cast   = 172 bytes
`` `

Rendered final evidence screenshots (produced by `--screen-renderer png`):
`final-evidence-BEFORE-origin-main.png` (blank) vs `final-evidence-AFTER-this-pr.png`
(readiness banner).

## 2. Narrowing the idle detector (review finding 1)

Target: draws `MAIN VIEW`, then emits an OSC terminal-title tick every 50 ms —
bytes keep arriving, rendered characters never change.
Recipe: `wait_for_text` then `wait_for_idle` (`stable_seconds: 0.5`, `timeout_seconds: 3`).

| session.py variant | result |
| --- | --- |
| raw byte length as ONGOING change detector (pre-review) | `FAIL 2:wait_for_idle - timed out waiting for idle` (0/1, burned the full 3s) |
| `current != last_screen or (stable_since is None and raw_len != last_raw_len)` (this PR) | `PASS 2:wait_for_idle - stable for 0.5s` (1/1) |

## 3. Silent target reports, rather than silently passing (review finding 2)

`termproof init initpack --name silent-daemon --command "sleep 30"` scaffolds a
pack whose only step is `wait for stable screen` / `wait_for_idle`. Running that
untouched scaffold against the silent target:

`` `
0/1 passed
FAIL silent-daemon [default]

### Steps
- FAIL `wait for stable screen` - no output observed from the session
`` `

versus the pre-existing wording, which is still used when output arrived but
never settled (`timed out waiting for idle`).

## 4. Uncapped idle cap cost is real (review finding 3, docs-only)

Silent long-lived target, no steps, recipe `timeout_seconds: 12`, so only the
post-script idle wait runs:

`` `
idle_cap_seconds: 3.0   -> wall  4.1s
idle_cap_seconds: null  -> wall 13.2s   (the whole recipe timeout, as documented)
`` `

## 5. Regression tests fail without the production change

`` `
origin/main session.py:
  FAIL test_wait_for_idle_does_not_report_idle_before_first_output  (True is not false)

screen-only arming (no raw-byte arm):
  FAIL test_wait_for_idle_arms_on_output_that_never_renders         (False is not true)

raw length as ongoing detector (un-narrowed):
  FAIL test_wait_for_idle_ignores_output_that_does_not_change_the_screen (False is not true)

this PR: all 6 QuiescenceBehaviorTest + 3 WaitForIdleStepDetailTest tests OK,
         5 consecutive repeats of QuiescenceBehaviorTest all OK (no flakiness).
`` `
```
</details>
<details>
<summary>Evidence: Slow-start daemon: TermProof report, before vs after</summary>

```text
BEFORE (origin/main termproof/session.py)
0/1 passed — FAIL slow-start-daemon [default]
Assertions: FAIL output_contains - contains 'SERVICE READY on port 8080'
Steps: PASS 1:wait_for_idle - stable for 0.5s <- armed on the blank startup screen
final.txt = 1 byte (empty), session.cast = 123 bytes (header only)

AFTER (this PR)
1/1 passed — PASS slow-start-daemon [default]
Assertions: PASS output_contains - contains 'SERVICE READY on port 8080'
Steps: PASS 1:wait_for_idle - stable for 0.5s <- armed at the first byte
final.txt = 27 bytes -> "SERVICE READY on port 8080", session.cast = 172 bytes
```
</details>
<details>
<summary>Evidence: Silent target through the untouched `termproof init` scaffold</summary>

```text
$ termproof init initpack --name silent-daemon --command "sleep 30"
created recipe: initpack/silent-daemon.recipe.json (only step: wait_for_idle, stable_seconds 0.75)

$ termproof run initpack/silent-daemon.recipe.json --no-video
0/1 passed — FAIL silent-daemon [default]
Assertions: PASS python traceback is absent
Steps: FAIL wait for stable screen - no output observed from the session
```
</details>
<details>
<summary>Evidence: Idle detector narrowing: OSC-title-ticking TUI, un-narrowed vs this PR</summary>

```text
raw byte length as ONGOING change detector (pre-review):
0/1 passed — PASS 1:wait_for_text - found 'MAIN VIEW'
FAIL 2:wait_for_idle - timed out waiting for idle (burned the full 3s)

current != last_screen or (stable_since is None and raw_len != last_raw_len) (this PR):
1/1 passed — PASS 1:wait_for_text - found 'MAIN VIEW'
PASS 2:wait_for_idle - stable for 0.5s
```
</details>
<details>
<summary>Evidence: Documented uncapped-timeout cost with a silent target (recipe timeout_seconds=12)</summary>

```text
idle_cap_seconds: 3.0 -> wall 4.1s
idle_cap_seconds: null -> wall 13.2s (the whole recipe timeout, exactly as the README states)
```
</details>

## Pipeline

Updates from [git push no-mistakes](https://github.com/kunchenguid/no-mistakes)

<details>
<summary>✅ **intent** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>⏭️ **Rebase** - skipped</summary>

- ⚠️ `termproof/session.py` - merge conflict rebasing onto origin/fm/flaky-pty-p4
- ⚠️ `tests/test_runner.py` - merge conflict rebasing onto origin/fm/flaky-pty-p4

</details>

<details>
<summary>⚠️ **Review** - 2 infos</summary>

- ⚠️ `termproof/session.py:129` - The raw-length *arming* half of the new condition is not pinned by any test. Walk the five QuiescenceBehaviorTest cases: `returns_true_when_screen_stabilizes` arms via the &#39;hello&#39; screen change; `times_out_when_output_never_stable`, `already_quiet`, and `ignores_output_that_does_not_change_the_screen` all arm from the entry seed (`stable_since = now if self.raw_output`) because each is gated behind a preceding `wait_for_text`; `does_not_report_idle_before_first_output` never arms at all and asserts False either way. Deleting `or (stable_since is None and raw_len != last_raw_len)` outright therefore leaves the entire suite green, even though the intent states it is deliberate and load-bearing (&#34;a first byte that is pure escape sequence never changes screen_text, so the timer would never start&#34;). Add a case whose child&#39;s only emission is escape-only — e.g. writes `\x1b]0;t\x07` once, then sleeps — entered without a preceding wait_for_text, asserting `wait_for_idle(0.3, 2.0)` is True; it fails (hangs to timeout) if the arming clause is removed.
- ℹ️ `README.md:232` - The README says the first-byte arming means &#34;a still-starting target is never captured as a blank &#39;idle&#39; screen&#34; (CHANGELOG.md:12 makes the same claim). The code guarantees something narrower: a target that has emitted *nothing* is never treated as idle. A target whose first bytes are an escape-only prologue — alt-screen enter `\x1b[?1049h`, or a `\x1b[2J\x1b[H` clear — arms the window immediately at session.py:129 while `screen_text` is still empty, so if its first rendered characters land more than `stable_seconds` later, the blank screen is still recorded. That is exactly the residual premature-idle race the intent scopes out, so this is a wording fix, not a code fix: state the guarantee as &#34;a session that has produced no output is never treated as idle&#34; rather than as a claim about still-starting targets generally.
- ℹ️ `termproof/session.py:131` - `last_raw_len = raw_len` is a dead store. The only reader of `last_raw_len` is guarded by `stable_since is None`, and every path that reaches this assignment also sets `stable_since` on the next line — so once written it can never be read. Since `stable_since is None` additionally implies `raw_output` was empty at entry, `last_raw_len` is always 0 when compared, and the clause reduces to `stable_since is None and self.raw_output`. Flagging as no-op rather than a fix: this is the exact expression the human reviewer verified, and churning it for two lines of cosmetics is not worth re-opening a hot spot.
- ℹ️ `termproof/runner.py:243` - Noting the accepted tradeoff for the record: the post-script `wait_for_idle(0.5, idle_timeout)` previously returned after ~0.5s for a silent, still-alive target and now burns the full budget — 3.0s under the default `idle_cap_seconds`, or the entire recipe `timeout_seconds` when the cap is null. This affects every recipe without `expect_exit_code`, not just `wait_for_idle` steps. It is deliberate per intent item 3, is literally consistent with the documented semantics of `null`, and is now covered by the README:232 sentence, so no action — recorded only so the cost is visible outside the PR description.

🔧 Fix: pin idle arming with a test; narrow docs guarantee
2 infos still open:
- ℹ️ `tests/test_runner.py:430` - The new arming test is the tightest ungated startup window in the suite. Because nothing renders, it cannot use a `wait_for_text` gate, so its single 2.0s budget must cover process startup (asciinema -&gt; shell -&gt; python) *and* the 0.3s stable window, leaving ~1.7s of startup slack. AGENTS.md:11-14, added in this same change, documents ~0.35s time-to-first-PTY-byte on a cold interpreter, and the sibling ungated case at tests/test_runner.py:334 allows 2.8s (3.0 timeout - 0.2 stable). On a loaded CI runner a first byte later than 1.7s flips this to `False` and the test fails for a timing reason, not a behavioural one - on a branch that exists specifically because of this class of flake. Raising the timeout to 5.0 costs nothing on the passing path (the call still returns at ~startup+0.3s) and does not weaken the assertion: without the arming clause the call still runs to the full deadline and returns False.
- ℹ️ `rust/crates/termproof-core/src/steps.rs:157` - Recording a new, narrower parity gap for the RUST-0xx workstream, not asking for a fix here. The intent already scopes out the Rust port&#39;s identical arming bug, but this change also introduces a second divergence the tracked item does not cover: the Rust `wait_for_idle` step emits only `timed out waiting for idle`, while Python now distinguishes `no output observed from the session`. rust/crates/termproof-cli/tests/cli_parity.rs does not assert step details, so nothing breaks today and CI stays green; the divergence would surface at cutover when the Rust backend starts producing user-visible step details that no longer match the README:232 contract.

</details>

<details>
<summary>✅ **Test** - passed</summary>

✅ No issues found.
- <code>`.venv/bin/python -m unittest tests.test_runner.QuiescenceBehaviorTest tests.test_wait_for_idle_step -v` (9 tests, OK; run with `.venv/bin` on PATH so the asciinema CLI resolves)</code>
- <code>`.venv/bin/python -m unittest tests.test_session tests.test_runner` (43 tests, OK)</code>
- <code>`.venv/bin/python -m unittest tests.test_docs_pages tests.test_release_docs tests.test_community_health tests.test_scaffold` (57 tests, OK — covers the CHANGELOG/README/AGENTS.md and init-scaffold surfaces this change touches)</code>
- <code>Flakiness probe: `tests.test_runner.QuiescenceBehaviorTest` run 5 consecutive times, all OK</code>
- <code>Fail-before/pass-after per test: with `termproof/session.py` from 080a3b2 → `test_wait_for_idle_does_not_report_idle_before_first_output` FAILs; with a screen-only arming variant → `test_wait_for_idle_arms_on_output_that_never_renders` FAILs; with raw length as an ongoing detector → `test_wait_for_idle_ignores_output_that_does_not_change_the_screen` FAILs (session.py restored, `git status` clean)</code>
- <code>E2E slow-start daemon via CLI: `termproof run recipes/slow_start_daemon.recipe.json --no-video` — origin/main session.py gives 0/1 passed, 1-byte final.txt, 123-byte cast; this PR gives 1/1 passed, final.txt &#39;SERVICE READY on port 8080&#39;, 172-byte cast</code>
- <code>E2E rendered evidence: same recipe re-run with `--screen-renderer png` before/after to capture the blank vs banner final screenshot</code>
- <code>E2E narrowing: `termproof run recipes/osc_title_tui.recipe.json` — PASS `2:wait_for_idle - stable for 0.5s` with the PR vs FAIL `timed out waiting for idle` with raw length as an ongoing detector</code>
- <code>E2E silent target through the scaffold: `termproof init initpack --name silent-daemon --command &#34;sleep 30&#34;` then `termproof run initpack/silent-daemon.recipe.json` → FAIL `wait for stable screen - no output observed from the session`</code>
- <code>E2E idle-cap cost: `termproof run recipes/silent_nostep.recipe.json --config config-{capped,uncapped}.yaml` timed — 4.1s with `idle_cap_seconds: 3.0` vs 13.2s with `null` against a recipe `timeout_seconds: 12`</code>
- <code>`git check-ignore -v .audit/scratch.txt` → matched by `.gitignore:16:.audit/` (scratch dir removed afterwards)</code>

</details>

<details>
<summary>⚠️ **Document** - 1 info</summary>

- ℹ️ `docs/rust-reimplementation-spec.md:131` - The Rust parity spec (and rust/crates/termproof-terminal/src/idle.rs doc comments) still describe wait_for_idle purely as screen-stability, with no record of the new first-byte arming invariant or the &#39;no output observed from the session&#39; failure detail; the Rust implementation retains the pre-change semantics. Deliberately left alone per the change&#39;s stated scope (tracked in the RUST-0xx parity workstream) — noting it so the invariant gets captured in the parity inventory when that workstream picks it up.

</details>

<details>
<summary>✅ **Lint** - passed</summary>

✅ No issues found.

</details>

<details>
<summary>✅ **Push** - passed</summary>

✅ No issues found.

</details>


md-mt:
## Review verdict

**VERDICT: CHANGES NEEDED**
**MERGE RISK: medium** — the fix is correct for the bug it targets, but it also silently redefines "idle" from *screen stable* to *screen stable AND no bytes*. Any TUI that emits bytes without changing rendered text can never satisfy a `wait_for_idle` step and will fail after burning the full timeout. A one-line narrowing fixes this, verified below.
**CI:** 11/11 green and meaningful — real PTYs on py3.11/3.12/3.13 plus an end-to-end evidence job. The headline regression test was confirmed to fail deterministically without the production change (5/5).
**Staleness:** clean. Merges onto current `main` (`080a3b2`) with no conflicts. No other open PR touches `session.py` or `tests/test_runner.py`.

The diagnosis is right, the regression tests are real, and the product bug was reproduced on current `main` in both directions. Two things block merge as-is.

### 1. major — `wait_for_idle` no longer returns for output that doesn't change the screen

`termproof/session.py:124-128`

The new detector is `current != last_screen or raw_len != last_raw_len`. `screen_text()` builds from `pyte.Screen.display`, which is **characters only** — no colours, no attributes — and pyte discards OSC/title sequences entirely. So a large class of normal TUI output produces a constant `screen_text` alongside a continuous byte stream. Those sessions were idle before; now they never are.

Measured against current `main`, `wait_for_idle(0.75, 3.0)`, with `s.screen` byte-identical throughout in every case:

| target | pre-fix | post-fix |
|---|---|---|
| repaints nothing, refreshes terminal title `\x1b]0;…\x07` every 100 ms | `True` @ 0.78 s | **`False` @ 3.02 s** |
| colour-cycling text, screen characters constant | `True` @ 0.78 s | **`False` @ 3.04 s** |
| idempotent full redraw `\x1b[H\x1b[2J…` at 10 Hz | `True` @ 0.31 s | **`False` @ 3.05 s** |

`README.md:223-227` documents this as waiting for "the **screen** to quiesce". The PR changes that contract without touching README or CHANGELOG.

**Suggested fix** — use raw length only to *arm* the timer, not as an ongoing detector:

```python
if current != last_screen or (stable_since is None and raw_len != last_raw_len):
```

Applied as a scratch experiment and re-verified: all 4 `QuiescenceBehaviorTest` tests OK; the OSC-title and colour-only cases restored to `True` @ ~0.78 s; silent-alive targets still `False`; and the PR's own slow-start product scenario still PASSes with `final.txt` = `SERVICE READY on port 8080`. The startup fix is preserved.

### 2. major — a silent-but-alive target flips PASS → FAIL, including the product's own scaffolded recipe

`termproof/session.py:120,129`; `termproof/scaffold.py:42-49`

A command that stays alive and emits zero bytes can never report idle. `termproof init` makes `wait_for_idle` the **only** step of every scaffolded recipe, so this hits the first thing a new user runs. Reproduced end to end over `python3 -c "import time; time.sleep(20)"`:

```
pre-fix  : 1/1 passed — PASS `wait for stable screen` - stable for 0.75s          (2.06 s)
post-fix : 0/1 passed — FAIL `wait for stable screen` - timed out waiting for idle (13.8 s)
```

Same for `cat` waiting on stdin. `asciinema rec --quiet` was confirmed to contribute zero bytes of its own, so there is no other signal to distinguish "still starting" from "silent by design".

This may well be the intended trade — for an evidence-recording tool, a zero-output target arguably *should* fail. But it is a user-visible change to the product's own generated recipe with no CHANGELOG entry, no README note, and a failure message that misdescribes what happened. **Suggested:** add a `### Changed` CHANGELOG entry and distinguish the detail string, e.g. `"no output observed from the session"` vs `"timed out waiting for idle"`.

No shipped example depends on the old behaviour — every `*.recipe.json` under `examples/`, `templates/` and `plugin-template/` was inventoried, and `termproof demo` passes end to end on the merged tree.

### 3. minor — `idle_cap_seconds: null` plus a silent target burns the whole recipe timeout

`runner.py:237-243`, `config.py:72-74`. Measured with `timeout_seconds: 12`: PASS in 1.3 s pre-fix, 13.0 s post-fix. This is literally consistent with the documented semantics, but the documented *intent* is "give slow-quiescing TUIs more room", not "always pay the full timeout for quiet targets". Worth a sentence in the README config block.

### 4. minor — the new `AGENTS.md` states a causal mechanism that does not hold

`AGENTS.md:5-11` asserts that "pytest's extra harness startup shifts PTY timing enough to expose races unittest never reaches", and instructs all future contributors to run PTY tests under both runners. Tested directly:

| setup (base code + base test) | result |
|---|---|
| `uv run python -m unittest …` | 10/10 pass |
| `uv run --with pytest pytest …` | 10/10 fail |
| `uv pip install pytest`, then the identical command | 6/6 pass |
| `uv pip uninstall pytest`, same command again | 6/6 fail |

The trigger is not pytest. It is `uv run --with pytest` building an **ephemeral overlay interpreter**, recreated per invocation. Time-to-first-PTY-byte there is 0.347 s cold versus 0.118 s in the warm project venv; the test uses `stable_seconds = 0.3`, so only the cold first session crosses the line.

Also: the "Regression example" the doc cites fails 5/5 under plain `unittest` too, so it is not a pytest-only failure; and pytest is not in the dev dependency group and CI never invokes it, so the doc prescribes a gate the project cannot enforce.

**Suggested:** keep the durable lesson, drop the mechanism — *"`wait_for_idle` must not arm its stable window before the child's first byte. A PTY test must never depend on first output arriving within `stable_seconds`; a cold interpreter or loaded machine can push it past 0.3 s."*

### 5. nit — the fix narrows the premature-idle race rather than closing it

A target that emits a banner, goes quiet for 2 s, then draws its main screen still reports idle during the gap — identical before and after. Inherent to timer-based idle detection, not a defect of this change, but worth stating so the fix is not over-credited.

### 6. nit — three unrelated changes in one diff

`.gitignore` and the new `AGENTS.md` are unrelated to the two-line behavioural fix in `session.py`, which would be far easier to review, revert and bisect on its own.

## What is good

- **The root cause is the actual root cause.** Verified rather than assumed: `asciinema rec --quiet` emits zero bytes of its own, so a still-starting child leaves `raw_output` genuinely empty. The diagnosis is precise.
- **The regression tests are real.** Reverting only `session.py` makes the new test fail 5/5 under CI's own runner.
- **The author proved the load-bearing claim and it holds.** That the *source* fix, not the added `wait_for_text`, resolves the flake: fixed `session.py` with the **unmodified** base test passes 10/10 where base code plus base test fails 10/10. That is exactly the experiment a skeptical reviewer would ask for.
- **The round-1 regression is genuinely pinned** — patching line 120 back makes `test_wait_for_idle_returns_true_for_already_quiet_session` fail while the others still pass. Both directions are held.
- **The product bug is real**, reproduced through the actual CLI on `main`: empty `final.txt` and a header-only cast before, correct output after. Not a test-only fix.
- **The new timing test is structurally safe, not lucky.** Its 0.6 s deadline against a 1.0 s sleep is a floor rather than an average, so it cannot flake in the false-pass direction.
- Using raw length as the *arming* signal is a subtle call the author got right — a first byte that is pure escape sequence would never change `screen_text`. Only its ongoing use overreaches.

PR #164, opened 2026-08-16
Attributed screen rendering: colour in the final screenshot and video, plus tmux and rsvg backends

md-mt:
## What you will see

**`final.svg` and the `attributed_rsvg` video are in colour.** The library used
to flatten pyte's buffer to plain text before any renderer ran, so foreground,
background, bold, italic, underline, strikethrough and reverse were all
discarded before a renderer could draw them. `termproof/attributed.py` keeps a
per-cell grid instead, and those two artifacts render from it.

**Per-step screenshots are not included, and that is most of them.** The images
under `steps/` render from `StepResult.screen`, which is pyte's already-flattened
`display`, so they stay monochrome — 124 of the 146 regenerated corpus SVGs.
Closing it needs `StepResult` to carry a grid, and an answer for how the corpus
pins those images.

**Dim (SGR 2) does not survive the cast-replay path.** pyte 0.8.2's `Char` has no
dim/faint field, so the attribute is gone before the grid is built. A grid parsed
straight from SGR text does carry it. Both limits are pinned by tests rather than
left to be found.

**The SVG markup shape changed.** One `<text>` per cell positioned at
`x = col * cell_w`, instead of one `<text>` per line. Column alignment is now
structural rather than dependent on whichever font the viewer resolves. Canvas
geometry is unchanged — for the default 80x24 the old and new formulas both give
756x516 — so anything sized against the artifact is unaffected.

Same recorded session, both renderers:

| | File | Distinct fill colours |
| --- | --- | --- |
| Before | [`examples/colorstress/before-monochrome.svg`](https://github.com/md-mt/termproof/blob/consolidate/attributed-rendering/examples/colorstress/before-monochrome.svg) | 1 |
| After | [`examples/artifacts/colour-stress/final.svg`](https://github.com/md-mt/termproof/blob/consolidate/attributed-rendering/examples/artifacts/colour-stress/final.svg) | 400 |

Both are checked into this branch and render in GitHub's file view.

## Two changes that are not additive

**The default session backend changed.** `pexpect` is now the default and writes
the asciinema v2 cast itself from the PTY output it is already reading. The
`asciinema` CLI is no longer required, and has moved out of the base
dependencies into a `termproof[record]` extra. Anyone who needs the cast to be
one asciinema itself wrote should install the extra and set
`session_backend: pexpect_asciinema`.

**`verify_provenance` is stricter.** It previously accepted any commit at all,
including the working tree's, which says nothing about what was tested. It now
requires `build_target`, `source_ref` and a binary that exists. Provenance that
used to verify may now not, which is the point.

## Also in here

- `png_rsvg`, a PNG renderer that rasterizes the same SVG the `svg` renderer
  emits, so the two cannot drift. Needs `librsvg`; `png` remains the default.
- `attributed_rsvg`, a video backend rendering frames from the same grid
  `final.svg` uses — not the per-step screenshots, which stay monochrome.
  `agg_ffmpeg` remains the default.
- A `tmux` session backend, so the screen comes from a real emulator's grid
  rather than a second emulator's reconstruction of the byte stream.
- `termproof.selection`, path-glob recipe selection.
- Step-screenshot dedup fingerprints the grid rather than the screen text. Note
  that `StepResult.screen` carries no escapes today, so the colour-only branch
  of that fingerprint cannot fire on a real run yet; its tests are labelled
  helper coverage.
- `render_attributed`, a new **optional** renderer method. Additive: renderers
  written against the text-only protocol keep working, which is pinned by a
  test. Documented in `docs/plugin-protocols.md`.

Full detail: [`CHANGELOG.md` under `## [0.3.0]`](https://github.com/md-mt/termproof/blob/consolidate/attributed-rendering/CHANGELOG.md).

## The corpus

`examples/artifacts/` is regenerated: 146 SVGs, all of which differ because the
markup shape changed. **No session was re-recorded** — the `session.cast` files
are untouched and the only input that changed is the renderer. That
regeneration is its own commit (`eec1330`) containing those 146 files and
nothing else.

`CorpusByteIdentityTest` is kept. Its purpose moves from "extracting the
rendering parameters must not change the output" to "the renderer's output is
pinned to a checked-in corpus", and its docstring says so.

## How this was verified

The branch was authored on a machine that could run only the standard-library
subset — nothing beyond the standard library had ever been executed against it,
and `uv.lock` had never been regenerated. Everything below was run for the first
time on this pass.

Re-run and green:

- `ruff check .`, `mypy termproof`
- the full 593-test suite on **3.11, 3.12 and 3.13** (3.11 needed nothing 3.12
  did not), plus 3.14 locally; `coverage report` at 82%
- `scripts/run_stdlib_tests.py` with no third-party packages installed
- the `verify` job's real evidence pipeline: `ci_evidence run ci`, 7/7 passing
  with video
- the plugin-template suite, config-wiring check and demo end-to-end
- the `tmux` backend on macOS with tmux 3.5a, unit tests and a full recipe
- `png_rsvg` and `attributed_rsvg` on a host with **no** `rsvg-convert` — both
  name the missing tool and the alternative

What this pass found and fixed, all of it code being wrong rather than tests:

- **the tmux backend recorded casts with every carriage return stripped.** The
  `pipe-pane` fifo was read in Python's default text mode, whose
  universal-newline translation rewrites `\r\n` and bare `\r` to `\n`. The
  recording looked complete; the replay was a diagonal staircase.
- **the corpus gate pinned a rendering path the pipeline does not use** — it
  rendered `final.svg` from text, where the pipeline renders it from the grid.
  Regenerating against it would have checked in monochrome screenshots the
  product never writes.
- **`attributed_rsvg` ignored `last_frame_duration`**, so the closing screen —
  the state the run ended in — showed for 42ms.
- **the wheel packaging gate could not survive a version bump**, and `uv.lock`
  still described 0.2.1.

Not verified, and not claimed: `attributed_rsvg` against real `rsvg-convert`.
No rasterizer would install or run in this environment — Homebrew, cairo and
headless Chrome each failed for a different reason. The backend was run end to
end with a stand-in rasterizer, which exercises frame generation, numbering,
the temp-directory lifecycle and the real ffmpeg encode; the resulting MP4
decodes cleanly at yuv444p/24fps. librsvg's own SVG parsing is the one step
that has not been exercised.

## Review round 1

Both findings on the first review were correct and are fixed here; neither was
something I had reported.

1. **Dim was claimed but dropped on the real path.** `_cell_from_pyte_char`
   never set `dim`, because pyte 0.8.2 has no such field — verified against the
   checked-in colour-stress cast, which carries two `ESC[2m` sequences while its
   `final.svg` carries zero `opacity` attributes. Claims narrowed; pinned by
   `test_dim_does_not_survive_the_cast_replay_path`, plus a companion test
   showing the SGR-text path does carry dim, so the limit is attributed to pyte
   rather than to the model.
2. **"Colour in every artifact" was wrong for most artifacts.** Narrowed the
   title, README, `docs/plugins.md`, the changelog and the fixture README to
   `final.svg` plus `attributed_rsvg`, excluding `steps/` explicitly. The
   synthetic-ANSI dedup tests are relabelled as helper coverage of
   `_render_step_screens`, and `test_a_real_step_screen_carries_no_escapes` now
   pins what a real step screenshot actually contains.

`AGENTS.md` carries both limits as a rule, so the claim is harder to rewrite by
accident.

## Review round 2

Both findings correct again, and the first is the more serious of the two.

1. **The dim guard could not fail.** It asserted on the `AttributedCell`, which
   `_cell_from_pyte_char` never populates with `dim` — so it read `False`
   regardless of what the emulator did. A guard against a claim going silently
   stale, which was itself incapable of detecting that. It now asserts on the
   raw pyte `Char`, where the fact actually lives. Verified by simulating a
   future pyte whose `Char` exposes `dim`: the guard fails with a message naming
   what to change. The consequence test is kept and relabelled as a consequence,
   not a guard.
2. **The old claim survived in places the first sweep did not reach.**
   `cast_video.py`'s module docstring, the `attributed_rsvg` bullet in
   `CHANGELOG.md`, and this PR body. Fixed all three, and swept every module
   docstring in `termproof/` programmatically rather than by grep this time —
   `attributed.py` had the same problem and is also fixed. `rsvg.py` was checked
   and makes no artifact-scope claim.

Sweep widened to every tracked surface rather than only the ones a round names,
which turned up a third survivor nobody had flagged: `site/index.html`, a
published page, still said the session is recorded "with asciinema" in both its
meta description and its lead paragraph. Fixed. Checked and clean this pass:
`docs-site/**`, `docs/guides/**`, `docs/case-studies/**`, `docs/releases.md`,
`docs/verified-badge.md`, `templates/`, `plugin-template/`, and every test
docstring. `_site/` is gitignored build output, not a surface.

**Every guard added on this branch has now been shown to fail when the thing it
guards is broken, then restored:** tmux CR preservation, corpus byte-identity
under a colour regression, the non-monochrome corpus check, the real-step-screen
pin, the closing-frame hold, the text-only renderer fallback, the stdlib-only
import gate, and the dim precondition.

## Review round 3

Both findings were on `site/index.html`, and the first was self-inflicted:
fixing the asciinema wording there in round 2, I wrote "colour SVGs and MP4s"
into the hero — reintroducing on the published page the exact blanket claim
rounds 1 and 2 were spent narrowing. Dropped the colour word from both summaries
rather than qualifying it. The second was a partial fix reading as a complete
one: I had grepped `with asciinema`, which matched two lines and not the third.

Rather than patch two more lines, the sweep is now a test.
`tests/test_public_claims.py` enumerates tracked claim surfaces from
`git ls-files` and fails on phrasings that have had to be withdrawn. On its
first run it found a fourth survivor nobody had reported —
`examples/apps/multi_turn_conversation.py` prints "Recording terminal session
with asciinema rec.", and that text is published in two corpus `final.txt`
files. Proved it bites on HTML, a Python module, markdown and package metadata.

**Surfaces enumerated** (so the next round can check the list, not rediscover
it): all `*.md`; `site/*.html` including `<meta>` and social tags; every `*.py`
including module docstrings; `pyproject.toml` description and keywords;
`Formula/termproof.rb`; `action.yml`; `docker/*Dockerfile`;
`docs-site/.vitepress/config.mts`; `examples/apps/*.py`, whose printed output
becomes corpus evidence; `*.sh`, `*.json`, `*.yml`. Deliberately excluded:
`examples/artifacts/**` and `site/artifacts/**` are recordings, where a stale
claim is a fact about when it was recorded; `docs/evidence/*.png` cannot be
linted and need an eye. The PR body is not tracked, so it stays manual.


md-mt:
VERDICT: request-changes
MERGE RISK: medium — the main renderer paths are well covered, but the PR still overstates artifact fidelity in two user-visible ways.

FINDINGS
1. termproof/attributed.py:386 — SGR 2/dim is still dropped on the actual final-screenshot/video path. `_cell_from_pyte_char()` copies the attributes pyte exposes, but pyte 0.8.2 has no `dim`/`faint` field; `screen_svg()` only emits `opacity="0.65"` when `cell.dim` is already true. I verified a minimal recorded cast containing `\x1b[2mdim\x1b[0m` replays through `replay_cast_attributed()` as `AttributedCell(... dim=False ...)`, and `examples/artifacts/colour-stress/session.cast` contains `\x1b[2m` while `examples/artifacts/colour-stress/final.svg` contains zero `opacity="0.65"` entries. That contradicts the user-facing "bold, dim, italic..." support in CHANGELOG.md:13 and examples/colorstress/README.md:13. Fix by either carrying dim through the cast replay path, or by removing dim from the supported-artifact claims and adding a product-path test that pins it as a known limitation.
2. README.md:295 — The public wording still says `attributed_rsvg` renders from "the same attributed grid the screenshots use" and that a video frame and screenshot of the same moment are the same image; docs/plugins.md:64 says the same thing, and the PR title says "colour in every artifact". But `termproof/evidence.py:167` writes step text from `StepResult.screen`, and `termproof/evidence.py:173` reparses that plain text for step screenshots. A real `examples/colorstress` run produced `final.svg` with 400 distinct fill colours, while `steps/*.svg` had only the default foreground/background fills. Also, 124 of the 146 regenerated corpus SVGs are step screenshots, so the title is not accurate for most rendered SVG artifacts. Fix by narrowing the title/docs to `final.svg` plus `attributed_rsvg`, explicitly excluding per-step screenshots until `StepResult` carries an attributed grid. The synthetic ANSI step-dedup tests at tests/test_evidence_config.py:444 are fine as helper coverage, but they should not be presented as pipeline coverage for real step screenshots.

CHECKED AND CLEAN
- tmux CR preservation: `_drain_pipe()` reads the `pipe-pane` fifo with `newline=""`; the real tmux tests passed, including `test_the_cast_keeps_carriage_returns` and the startup gate test.
- Corpus gate path: `CorpusByteIdentityTest` now renders final SVGs through attributed cast replay and step SVGs from recorded step text, matching `render_artifacts`; the corpus byte-identity test passed.
- `attributed_rsvg` last-frame hold: `agg --help` reports `--last-frame-duration` default `3`; the backend default is 3.0s, and held frames are copied rather than re-rasterized.
- `uv.lock` is at `termproof` 0.3.0 and keeps `asciinema` behind the `record` extra; the sdist/wheel artifact-content gate passed.
- Text-only third-party renderers still work: helper and whole-pipeline fallback tests passed.
- `scripts/run_stdlib_tests.py` is the first CI job; baseline passed under `/usr/bin/python3`, and a temporary local `import pyte` in `attributed.py` made it fail with `ModuleNotFoundError` as claimed, then passed again after the edit was removed.
- Control characters are dropped both during grid construction and glyph emission, with XML well-formedness tests covering the path.
- The default session backend and stricter provenance checks are called out in the changelog/README; targeted provenance and internal-recorder tests passed.
- Python 3.11 spot-check passed for the new modules/tests (`test_attributed`, `test_cast_video`, `test_rsvg`, `test_evidence_config`, `test_build_info_provenance`, `test_selection`).
- Full suite passed once: `uv run python -m unittest discover -s tests` ran 593 tests successfully.
- No `rsvg-convert` is installed on this host, so real librsvg parsing remains unverified; missing-tool and nonzero-failure paths are tested and fail loudly as far as subprocess status is concerned.

md-mt:
VERDICT: request-changes
MERGE RISK: medium — Product behavior and most docs are now aligned, but one guard test and one added public docstring still leave the original overclaim failure mode open.

FINDINGS
1. tests/test_screen.py:51 — The dim regression test does not actually fail when pyte starts exposing `dim`/`faint`; it only checks the `AttributedCell` after `termproof/attributed.py:386` has already ignored any such attribute. I verified `_cell_from_pyte_char()` still returns `dim=False` for a char-like object with `dim=True` and `faint=True`, so a pyte upgrade could make SGR 2 available while the cast/video path keeps dropping it and this guard stays green. Fix by asserting the raw pyte char for an SGR 2 cell has no `dim`/`faint` field, or by carrying the new field through and updating the limitation docs.
2. termproof/cast_video.py:4 — The module docstring still says `attributed_rsvg` replays into "the same attributed screen the screenshots use" and that a video frame and "a screenshot" of the same moment are the same image. That is still the old claim for per-step screenshots: `_render_step_screens` renders from `StepResult.screen`, and the real colour-stress step SVGs contain only the default foreground/background fills. The PR body also repeats this wording in the `attributed_rsvg` bullet. Fix by narrowing both to `final.svg` / the final screenshot and explicitly excluding `steps/`, matching `README.md` and `docs/plugins.md`.

CHECKED AND CLEAN
- The PR title is now accurate: `final.svg` and the `attributed_rsvg` video are the colored artifacts, not every artifact.
- `README.md:295`, `docs/plugins.md:64`, `CHANGELOG.md:12`, and `examples/colorstress/README.md:13` now narrow the claim to `final.svg` plus `attributed_rsvg`, and exclude per-step screenshots / dim-on-cast replay.
- The real video path is colored: a recorded cast with SGR red went through `RsvgFfmpegBackend` using the real pyte replay, and the generated frame SVG contained `#ff7b72`.
- The real step path is still monochrome and honestly pinned: `test_a_real_step_screen_carries_no_escapes` drives `pyte.Screen` -> `screen_text` like the runner, and colour-stress step SVGs have only default fg/bg.
- The two synthetic ANSI step-dedup tests are visibly labelled helper coverage of `_render_step_screens`, not pipeline coverage.
- Corpus scale matches the narrowed claim: 23 final SVGs, 133 step SVGs; colour-stress `final.svg` has 400 distinct fills, while its step SVGs have only the two defaults.
- Actual dim behavior matches the narrowed docs today: the colour-stress cast has SGR 2, `final.svg` has no opacity, bold from the same cast survives, and the SGR-text render path does emit `opacity="0.65"`.
- Round-1 clean items stayed clean: tmux CR preservation/startup gate, product-path corpus gate, last-frame hold, lockfile / optional recorder dependency, text-only renderer fallback, stdlib CI job, control-character dropping, and 3.11 spot coverage.
- Validation run locally: `ruff check .`, `mypy termproof`, `python -m unittest discover -s tests` (596 tests), stdlib-only loader under `/usr/bin/python3`, focused review suites, and a 3.11 spot-check all passed.
- Added-line sweep found no new instances of the prohibited name categories.
- Real `rsvg-convert` / `ffmpeg` execution was not available on this host; the subprocess seams and generated SVG inputs were exercised with stand-ins.

md-mt:
VERDICT: request-changes
MERGE RISK: medium — the core implementation and guard coverage look solid, but the published site still has user-visible claims that contradict the narrowed behavior.

FINDINGS
1. site/index.html:27 — The updated hero says TermProof replays the cast into `colour SVGs and MP4s`, and the updated meta description at line 7 says `colour screenshots and MP4`. A stranger can read that as applying to the screenshot/SVG artifact set, but the PR body and code now explicitly say only `final.svg` and the `attributed_rsvg` video render from the attributed grid; per-step screenshots under `steps/` stay monochrome. This reopens the round-1 artifact-scope overclaim on the published page. Narrow this to `final.svg` plus the attributed video backend, or avoid the colour claim on the site summary.
2. site/index.html:42 — The same published page still says `<code>asciinema</code> records the session`. The PR changes the default backend so `pexpect` writes the asciinema v2 cast itself, with the asciinema CLI only behind `termproof[record]` / `pexpect_asciinema`. This means the third-survivor fix for `site/index.html` is still partial: the meta description and lead paragraph were changed, but this card keeps the old recorder claim. Replace it with wording about recording an asciinema-format cast internally.

CHECKED AND CLEAN
- Read both prior review comments and the current PR title/body. The title/body now narrow colour to `final.svg` and `attributed_rsvg`, explicitly exclude per-step screenshots, and describe the dim limitation.
- Mutation spot-checks all bit: simulating pyte with `dim` failed `test_the_emulator_still_exposes_no_dim_attribute`; neutering `cell_colors` foreground failed corpus byte identity on `examples/artifacts/colour-stress/final.svg`; forcing `render_attributed` failed the text-only renderer fallback tests. Each mutation was restored.
- Working tree is clean after restores at PR head `3c8ed9bd6bcc3bb574eb911461b87faefafa6471`.
- Focused regression suites passed: screen/corpus/dedup/fallback/cast-video tests (52), stdlib-only gate under `/usr/bin/python3` (35), and tmux/internal-recorder/provenance/rsvg tests (118, 1 skipped).
- `gh-axi pr checks 164` reports 12 passed, 0 failed, 2 skipped.
- Root README, CHANGELOG, docs/plugins, docs/evidence-quality, the colour-stress README, module docstrings, test docstrings, and CLI help strings use the narrowed artifact/dim wording outside the site findings above.
- Added-line exposure sweep found no new prohibited-name instance.

md-mt:
VERDICT: approve
MERGE RISK: low — The new guard covers the tracked surfaces that escaped earlier rounds, the current PR body is manually clean, and local plus remote validation is green.

FINDINGS
None.

CHECKED AND CLEAN
- Read the three prior review comments and the current full PR body; the current claims are narrowed to `final.svg` and `attributed_rsvg`, exclude per-step screenshots, and state the dim limitation.
- `tests/test_public_claims.py` bites on real tracked surfaces: temporary mutations failed for `site/index.html` (`colour screenshots`), `termproof/cast_video.py` (`the same attributed screen the screenshots use`, caught by both whole-file and docstring tests), `README.md` (`Colour in every artifact`), and `pyproject.toml` (`colour screenshots`). Each mutation was restored.
- The enumeration covers 226 tracked surfaces via `git ls-files`, including markdown, site HTML, Python/module docstrings, package metadata, workflows/config, Dockerfile, and example app output. Breaking the markdown glob made the enumeration guard fail on `README.md`.
- The banned list matches the actual historical survivor strings I checked: unsupported dim attribute lists, same-grid screenshot wording, same-image wording, site colour plurals, site recorder wording, and the demo app recorder line. The narrowed `colour screenshots` and `bold, dim,` patterns still catch the original defects without flagging the qualified replacement text.
- Exclusions look defensible: recorded evidence stays historical, non-text images need human review, and the PR body is outside `git ls-files`. I think manual PR-body checking is adequate for the PR discussion surface; release/user-facing claims should continue to live in tracked docs.
- The demo app now prints `Recording terminal session to an asciinema-format cast.`; the live multi-turn recipe passed with `--no-video`; the stale old line is confined to checked-in recorded evidence snapshots and is not asserted as current behavior.
- No new prohibited-name exposure found in the added review surface.
- Validation passed: `uv run python -m unittest tests.test_public_claims`, `uv run ruff check .`, `uv run mypy termproof`, `/usr/bin/python3 scripts/run_stdlib_tests.py` (118 tests, 1 skipped), `uv run python -m unittest discover -s tests` (600 tests), and remote checks at 12 pass / 2 skip.
- Working tree clean after all mutation restores.

PR #176, opened 2026-08-18
docs: correct the distribution claims that went stale when 0.3.4 shipped

md-mt:
**Scope: the prose corrections only.** Round 2 asked for the split and I agree with it. The guard and
the rewriter are out of this PR entirely — see [What the follow-up should
carry](#what-the-follow-up-should-carry) at the bottom.

## The defect

Both packages published at **0.3.4** — `termproof` on PyPI and the `termproof` crate on crates.io,
both verified live against the registry APIs. The repository still said otherwise in a dozen places.

| Surface | Said | Now |
| --- | --- | --- |
| `README.md:29` distribution table | "not yet on PyPI" | links the PyPI project |
| `README.md:238` what-is-published | PyPI "no — gated behind `ENABLE_PYPI`" | `yes, through 0.3.4` — the variable has been `true` since 2026-08-16 |
| `README.md:239` what-is-published | crate "yes, through 0.3.3" | `yes, through 0.3.4` |
| `README.md:178` | "Releases through 0.3.3 were cut before consolidation; from here they are tagged `rs-v*`" | true as history, but "from here" reads as *not yet* now that `rs-v0.3.4` exists |
| `README.md` install block | git-only | offers `pip install termproof` first |
| `SECURITY.md:59` | PyPI "no — gated"; a fix reaches consumers "once the gate is on" | `yes, through 0.3.4` |
| `SECURITY.md:60` | crate "through 0.3.3 (…enumerated…)" | `through 0.3.4`, enumeration replaced by a registry link |
| `rust/docs/publishing.md:3,17,25` | "currently through 0.3.3", "published through 0.3.3", "shipped (0.2.1 through 0.3.3)" | 0.3.4 |
| `rust/docs/publishing.md:170` | "releases through `v0.3.3` were cut from the separate repo" | history, reworded so it does not read as a currency claim |
| `rust/docs/publishing.md:262` | "it has not yet run on an `rs-v*` tag here, because none has been cut" | it ran on `rs-v0.3.4`; six assets are attached to that release |
| `rust/crates/termproof-cli/README.md:26` | "releases through `0.3.3`" | history, reworded |
| `rust/docker/termproof.Dockerfile:14` | "reach today (v0.3.3)" | v0.3.4 |
| `.github/workflows/rust-security.yml:72` | "termproof 0.3.3 today" | 0.3.4 |
| `.github/workflows/rust-release.yml:7` | "ran on every tag from v0.2.1 through v0.3.3 … has not yet run on an `rs-v*` tag" | through v0.3.4, naming `rs-v0.3.4` |
| `.github/workflows/rust-auto-release.yml:181` | "No `rs-v*` tag is reachable from `main`" | both `rs-baseline-v0.3.3` and `rs-v0.3.4` are; the guard is dormant and says so |
| `python/action.yml:5` | "requires a published release (v0.2.1+). Until then…" | it is published |
| `python/docs/ci/circleci.md:45` | "Use `termproof` after PyPI publishing" | how to use the published package, and how to pin |
| `python/tests/test_docs_pages.py:80,124` | *forbade* `pip install termproof` "(not on PyPI)" | requires it |
| `python/README.md:37` | git-only install on the page PyPI renders | offers `pip install termproof` |

Verified live, not assumed:

- PyPI `termproof`: 0.3.0, 0.3.3, **0.3.4** (uploaded 2026-08-17T06:31:50), none yanked.
- crates.io `termproof`: 0.2.1 → **0.3.4** (2026-08-17T06:59:38), none yanked.
- `termproof-cli` and `termproof-plugin-protocol`: 404 on crates.io. Those "not published" claims are
  still true and are left alone, as is `README.md:172`.
- Release (Rust) succeeded on `rs-v0.3.4`; the release carries three archives and three `.sha256`.
- `ENABLE_PYPI` has been `true` since 2026-08-16, and the `py-v0.3.4` release run succeeded.

## What is *not* in this PR

Reverted to `origin/main`, byte-identical (`git diff origin/main` is empty for both):

- **`.github/scripts/rust/version-bump.py`** — no claim scanner, no `rewrite_claims`, no
  per-artifact scoping, no `--publishes`.
- **`python/tests/test_current_version_claims.py`** — back to asserting agreement between surfaces.
- **`rust-auto-release.yml`'s `--publishes crates` argument.**

The manifest tie and the rewriter are entangled by construction: it is only safe to demand prose
match the manifest if something moves the prose in the bump commit. Without that it red-mains the
next release — which is the objection the file's original docstring raised, and it was right. So the
test goes with the rewriter rather than landing half of a mechanism.

**Nothing in the release path changed.** The three workflow files carry comment-only edits:

```
.github/workflows/rust-auto-release.yml       parsed content IDENTICAL
.github/workflows/rust-release.yml            parsed content IDENTICAL
.github/workflows/rust-security.yml           parsed content IDENTICAL
```

(`yaml.safe_load` of each, against `origin/main`.) And `version-bump.py 0.3.5 --check` still reports
exactly the four edits it reported before this branch existed.

## Two things kept, and why

**`rust/docs/publishing.md`.** Its correction stays, but the version-bump rule is rewritten for the
world where no automation moves the claims. Round 2 established two facts that hold regardless of
whether the rewriter ever lands, and the doc now says both: the manual step has to move *all* the
claims — the whole set went stale together after 0.3.4 because the step was skipped — and it has to
be scoped to what actually shipped, because a Rust release cuts `rs-v*` while PyPI is only uploaded
on `py-v*`. The existing sweep compares surfaces to each other, so it cannot see either failure, and
the doc says that too.

**`python/tests/test_docs_pages.py`.** It asserted that `pip install termproof` must be *absent*
because the package was unpublished. That is now false, it depends on nothing being reverted, and
leaving it would re-encode in a test the exact claim this PR exists to correct.

## Verification

| Check | Result |
| --- | --- |
| `ruff check .` | pass |
| `mypy termproof` | pass, 40 files |
| `scripts/check_version.py` | consistent at 0.3.4 |
| `python -m unittest discover -s tests` | **684 tests, OK** |
| `scripts/run_stdlib_tests.py` | 118 tests, OK |
| `scripts/validate_community_health.py` | all checks passed |
| `cargo fmt --check --all` | pass |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | pass |
| `cargo test --workspace` | 24 targets, zero failures |
| clippy + test, all 16 feature combinations | 16/16 pass |

The existing `test_current_version_claims.py` — the unmodified one — passes against the corrected
prose and finds all twelve current-version claims agreeing at 0.3.4:

```
.github/workflows/rust-release.yml:7      README.md:242         rust/docs/publishing.md:3
.github/workflows/rust-security.yml:72    README.md:243         rust/docs/publishing.md:18
rust/docker/termproof.Dockerfile:14       SECURITY.md:59        rust/docs/publishing.md:26
                                          SECURITY.md:60        rust/docs/publishing.md:272,273
```

## What the follow-up should carry

The guard and the rewriter, with these four unsolved. I am not opening it.

1. **The rewriter can corrupt text.** `Published through https://example.invalid/releases/0.3.3/
   today` rewrites the URL path. `Supported today: >=0.3.3` rewrites the version bound.
   `As of 2026.08.17` is read as a claim for version `2026.08.17` and would red-main a correct
   document. Three rounds of patching one regex is the signal — the follow-up should look at marking
   claims explicitly (an HTML comment, a fenced marker, a data file the docs are generated against)
   rather than pattern-matching them out of free prose.

2. **The Python release path never moves claims.** `python-release.yml` uploads to PyPI and never
   calls the rewriter, so the original bug is unfixed on that side. The scoped guard permits it,
   because the crates claim satisfies "some artifact matches the manifest". Adding the obvious
   `--publishes pypi` call fails the guard's own "the rewrite reached almost nothing" assertion,
   which assumes the crates side supplies the count. The design does not currently admit a
   Python-first release, and that has to be designed rather than patched.

3. **The denial check has a circular oracle.** "Recorded as published" is derived from the
   repository's own prose, so replacing both publication rows with denials makes the check stop
   checking PyPI at all. Same for a denial phrased outside `NOT_PUBLISHED`, and for one in an
   excluded directory. I documented the first as a limitation; that was too generous to it. The
   guard needs evidence from outside prose — a committed release receipt written by the release
   workflow is the shape I would try first, since it is offline at test time but produced by the
   thing that actually uploaded.

4. **Wrapped prose is narrowed, not solved** (non-blocking). A claim that wraps *with* a version on
   the first line is read as a claim for the wrong version and silently not rewritten —
   ``The releases that have shipped (`0.2.1` through`` / ``` `0.3.4`) carried the library only. ```
   reads as a claim for 0.2.1. Current files avoid the shape; the class remains.

## Out of scope, flagged not fixed

**Docker Image (Rust) has been failing** on `main` and on `rs-v0.3.4` with `denied:
permission_denied: write_package`, so `ghcr.io/md-mt/termproof-rust` has no `rs-v0.3.4` tag.
`SECURITY.md` says the images publish "on every push to `main` and every release tag" — a correct
description of the intended trigger that is not currently happening for the Rust image. That is a
broken build, not a stale sentence, so it belongs in its own change. (It passes on this PR.)

## Caveats preserved

Nothing near the per-step screenshots (monochrome), the conformance corpus (unstable baseline,
layer-level measurements) or the Rust CLI's unreached external tools was weakened.

---

# Review round 3 — one blocking, two non-blocking, all fixed

## Blocking: a stale embedded transcript, in a file this PR already touches

`rust/docs/publishing.md:63` pasted a `publish-plan.py` run printing
`"version": "0.3.3"`. The script prints `0.3.4`.

Bumping the number would have re-set the same trap, so the transcript is gone.
The page names the command and describes the shape of what it returns —
`version` from the manifests, `order`, `held` — and says why an embedded sample
of live tool output is a claim no test can check and no bump will move. The
"do not copy that anywhere" instruction now applies to the page itself.

## The `0.2.0` pin that never existed on PyPI

`python/docs/releases.md:96` and `python/scripts/smoke-install.sh:5` documented
`smoke-install.sh 0.2.0`. PyPI carries `0.3.0`, `0.3.3`, `0.3.4` — never
`0.2.0` — so the documented smoke test could not succeed and never could. Both
take a `<version>` placeholder now, rather than today's number, which would rot
identically.

`python/docs/launch/checklist.md:60` I handled differently, and deliberately.
Its `0.2.0` sits beside "Tag `v0.2.0` pushed" (line 57) and "Version bumped in
`pyproject.toml` to `0.2.0`" (line 48) in a plan for a launch that happened at a
different number; placeholdering one of the three would read worse than leaving
them. The header now states that the plan was never executed at `0.2.0`, that
the tag and GitHub release exist but nothing reached PyPI under that number,
and that the `pip install termproof==0.2.0` step therefore cannot succeed. Every
number below it then reads as plan rather than fact.

## The container-image rows

Both `README.md` and `SECURITY.md` said the two images publish "on every push to
`main` and every release tag". Verified against the registry:

| Image | Tags | Verdict |
| --- | --- | --- |
| `ghcr.io/md-mt/termproof` | `main`, `latest`, `v0.3.0`, `v0.3.3`, `py-v0.3.4` | claim holds |
| `ghcr.io/md-mt/termproof-rust` | `main`, `latest`, `v0.3.3` | claim false |

The Rust image's `latest` and `main` were both built `2026-08-16T19:22:36Z` from
revision `bb8dc3af` — the same build as `v0.3.3`. There is no `rs-v0.3.4` image.

They are separate rows now. The Rust one says it is stale, says `latest` is
still the `0.3.3` build, and points at #178; the README paragraph offering that
image carries the same note. **The build is not fixed here** — #178 owns it.

## Verification

| Check | Result |
| --- | --- |
| `ruff check .` | pass |
| `mypy termproof` | pass, 40 files |
| `scripts/check_version.py` | consistent at 0.3.4 |
| `python -m unittest discover -s tests` | **684 tests, OK** |
| `scripts/run_stdlib_tests.py` | 118 tests, OK |
| `scripts/validate_community_health.py` | all checks passed |
| `bash -n scripts/smoke-install.sh` | parses |
| `cargo fmt --check --all` | pass |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | pass |
| `cargo test --workspace` | 24 targets, zero failures |
| clippy + test, all 16 feature combinations | 16/16 pass |
| `version-bump.py`, `test_current_version_claims.py` | still byte-identical to `origin/main` |
| `publish-plan.py` output vs the page | page no longer asserts one |

### DIFF
--- a/.github/scripts/rust/version-bump.py
+++ b/.github/scripts/rust/version-bump.py
@@ -11,7 +11,7 @@ release cannot leave the Python manifest or the changelog behind — which is
 exactly what it used to do, bumping `rust/Cargo.toml` alone and pushing a
 `main` that failed its own drift check.
 
-Four places carry the version and must move together:
+Five places carry the version and must move together:
 
   - `rust/Cargo.toml`, `[workspace.package] version`, which every member
     inherits;
@@ -25,7 +25,15 @@ Four places carry the version and must move together:
     section is promoted to a heading for this version and a fresh empty
     `[Unreleased]` takes its place. Keep a Changelog's release step, and what
     keeps the curated entries contributors wrote rather than generating prose
-    over the top of them.
+    over the top of them;
+  - every *current-version claim* in tracked prose — the sentences in
+    `README.md`, `SECURITY.md`, `rust/docs/publishing.md`, workflow headers and
+    Dockerfile comments that say what is published *today*. These were the one
+    part of the train nothing moved, so they drifted a release behind and stayed
+    there until a human noticed. `CURRENCY` below is what makes a version a
+    claim about now rather than an example or a range's lower bound, and
+    `python/tests/test_current_version_claims.py` imports it from here so the
+    sweep and the rewriter cannot disagree about what counts.
 
 Internal dependencies are identified by the presence of `path = `, not by
 name. No crate is named in this file, so merging several crates into one, or
@@ -175,6 +183,104 @@ def promote_changelog(text, new, date):
     return updated, [f"CHANGELOG.md: [Unreleased] -> [{new}] \u2014 {date}"]
 
 
+#: Phrasings that make a version a statement about *now* rather than an
+#: example, a dependency pin or a historical range's lower bound.
+CURRENCY = re.compile(
+    r"\btoday\b|\bcurrently\b|\blatest published\b|\bVERIFIED\b|\bso far\b"
+    r"|\bas of\b|\bthrough\b|\bhas run\b|\bran\b",
+    re.IGNORECASE,
+)
+
+#: A bare or `v`-prefixed semver, not part of a longer dotted string (so
+#: `1.0.69` inside `thiserror 1.0.69` still matches, but `0.3.3.1` does not).
+CLAIM_VERSION = re.compile(r"(?<![\w.])v?(\d+\.\d+\.\d+)(?![\w.])")
+
+#: Generated output, vendored tooling, lockfiles, and the two files that have
+#: to spell out example claims in order to define or test what a claim is —
+#: the Python suite, and this script's own docstrings.
+CLAIM_EXCLUDED_PREFIXES = (
+    "python/examples/artifacts/",
+    "python/site/artifacts/",
+    "python/_site/",
+    "conformance/corpus/",
+    "rust/.specify/",
+    "rust/.claude/",
+    "python/.hermes/",
+    "python/tests/",
+    "python/docs-site/package-lock.json",
+    "python/uv.lock",
+    "rust/Cargo.lock",
+    ".github/scripts/rust/version-bump.py",
+)
+
+
+def _newest(versions):
+    return max(versions, key=lambda v: [int(part) for part in v.split(".")])
+
+
+def claim_windows(lines):
+    """Yield `(index, claimed_version, span)` for each current-version claim.
+
+    `span` is the line indices the claim occupies and `claimed_version` is the
+    newest version inside it, because a claim written as a range ("from
+    v0.2.1 through <the version>") is a claim about its upper bound.
+
+    A claim is one line, except when the sentence wrapped: a currency word with
+    no version after it on its own line is the first half of something like
+    `... every tag from v0.2.1 through` / `<version> ...`, and a line-only scan
+    reads the range's lower bound as the answer. That exact wrap is how a
+    workflow header escaped an earlier sweep. The extension is conditional
+    rather than unconditional because an unconditional second line swallows
+    whatever prose happens to follow a claim — which, when this rewrites, means
+    editing a sentence that was never a claim at all.
+    """
+    for index, line in enumerate(lines):
+        if not CURRENCY.search(line):
+            continue
+        span = [index]
+        versions = CLAIM_VERSION.findall(line)
+        if not versions and index + 1 < len(lines):
+            span.append(index + 1)
+            versions = CLAIM_VERSION.findall(lines[index + 1])
+        if versions:
+            yield index, _newest(versions), span
+
+
+def claim_files(repo):
+    """Tracked, repository-relative paths that may carry a claim."""
+    listing = subprocess.run(
+        ["git", "ls-files"],
+        cwd=repo,
+        check=True,
+        capture_output=True,
+        text=True,
+    ).stdout.split()
+    return [name for name in listing if not name.startswith(CLAIM_EXCLUDED_PREFIXES)]
+
+
+def rewrite_claims(text, old, new):
+    """Return (new_text, [edits]) with `old` -> `new` inside claims only.
+
+    Only lines belonging to a window whose claimed version is `old` are
+    touched, so examples, dependency pins and the lower bound of a historical
+    range keep the number they had. A `v` prefix is preserved.
+    """
+    lines = text.splitlines(keepends=True)
+    targets = set()
+    for _, claimed, span in claim_windows([line.rstrip("\n") for line in lines]):
+        if claimed == old:
+            targets.update(span)
+
+    pattern = re.compile(rf"(?<![\w.])(v?){re.escape(old)}(?![\w.])")
+    edits = []
+    for index in sorted(targets):
+        replaced, count = pattern.subn(rf"\g<1>{new}", lines[index])
+        if count:
+            lines[index] = replaced
+            edits.append(f"line {index + 1}: current-version claim -> {new}")
+    return "".join(lines), edits
+
+
 def read_pyproject_version(path):
     section = None
     for line in path.read_text(encoding="utf-8").splitlines():
@@ -268,6 +374,26 @@ def main():
     else:
         print(f"CHANGELOG.md already carries a heading for {new}")
 
+    # Claims last, and over the text the edits above produced rather than over
+    # what is still on disk — the changelog has already been promoted in
+    # memory by this point, and reading it back would undo that.
+    pending = dict(writes)
+    if old != new:
+        for name in claim_files(repo):
+            path = repo / name
+            if path in pending:
+                text = pending[path]
+            else:
+                try:
+                    text = path.read_text(encoding="utf-8")
+                except (UnicodeDecodeError, OSError):
+                    continue
+            text, claim_edits = rewrite_claims(text, old, new)
+            if claim_edits:
+                edits += [f"{name}: {edit}" for edit in claim_edits]
+                pending[path] = text
+    writes = list(pending.items())
+
     if not edits:
         print(f"already at {new}; nothing to do")
         return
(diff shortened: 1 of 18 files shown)


md-mt:
## Review summary

Three rounds. The first two were against a substantially larger change; this PR was narrowed in
response to them, and the third round reviewed what is actually here.

### Round 1 — do-not-merge (3 blocking)

Against the original change, which added a manifest-bound claims guard and a release-time prose
rewriter.

1. `CLAIM_VERSION` missed a version followed by sentence punctuation, so `0.3.3.` at the end of a
   sentence was invisible to both the sweep and the rewriter.
2. The `not yet on PyPI` row — the claim that prompted this work — was still unguarded. Reverting it
   passed all three guards.
3. **A Rust-only release could publish a false PyPI claim.** The rewriter moved claims for both
   registries, but PyPI is only uploaded on a `py-v*` tag while the Rust auto-release cuts `rs-v*`.
   Reproduced with a real bump: `README.md` and `SECURITY.md` were rewritten to assert a PyPI version
   that had not been published, with every check green.

### Round 2 — do-not-merge (3 blocking)

All three round-1 items were addressed, and the per-artifact scoping worked correctly for the Rust
path. New findings:

1. The Python release path never invoked the rewriter at all, so PyPI claims still went stale after a
   Python release. Adding the obvious call failed one of the guard's own assertions, which assumed
   the crates side supplied the version witness.
2. Remaining false positives in the matcher: it rewrote a version inside a URL path, rewrote a
   `>=0.3.3` version bound, and read a date-like `2026.08.17` as a version — the last of which would
   fail CI on a correct document.
3. The publication-denial check derived "is this registry published?" from the repository's own
   prose. Replacing the publication rows with denials made the check stop checking.

### Decision: split

The prose corrections were correct and the defect they fix was live and public. The guard had taken
three rounds and each round found a new way for it to damage a release — worse than the drift it
prevents. The guard and rewriter were removed to a follow-up; this PR keeps the corrections.

### Round 3 — merge, after three fixes

Verified against the narrowed change:

- `version-bump.py` and `test_current_version_claims.py` are byte-identical to `main`; no dangling
  references to the removed code.
- The three workflow edits are comment-only — `yaml.safe_load` is identical to `main` for each.
- Corrected claims match the registries: PyPI and crates.io both at 0.3.4; `termproof-cli` and
  `termproof-plugin-protocol` are correctly described as unpublished.
- The `test_docs_pages.py` inversion is right and complete; nothing else encodes the old
  "unpublished" assumption.
- Caveats intact: per-step screenshots still described as monochrome, the conformance corpus not
  described as stable, the CLI's external-tool claims accurate.
- 684 Python tests, 118 stdlib tests, `cargo test --workspace`, and all 16 feature combinations pass.

Fixes applied after round 3:

1. **Blocking** — `rust/docs/publishing.md` embedded a `publish-plan.py` transcript showing `0.3.3`;
   the script prints `0.3.4`. Fixed by not embedding live tool output, rather than by bumping the
   number.
2. Three files documented a smoke test pinned to `0.2.0`, a version never published to PyPI.
3. `SECURITY.md` claimed a container image on every release tag. Untrue for the Rust image — see
   #178, which is a separate publishing-permission failure, not a docs defect.

### Not covered by this PR

The follow-up guard, with four unsolved problems: the rewriter corrupting URLs, version bounds and
date-like strings; the Python release path; the denial check's circular oracle; and wrapped prose.
The suggested direction is to mark claims explicitly rather than pattern-match them out of free
prose, and to source publication evidence from a release receipt rather than from prose.

PR #181, opened 2026-08-19
Carry the attributed grid to the per-step screenshots

### LINKED ISSUE #175: Per-step screenshots miss the attributed rendering because StepResult.screen is flattened text
Attributed rendering reaches `final.svg` and the `attributed_rsvg` video, but
not the per-step screenshots: `StepResult.screen` is a `str`, so the grid is
gone before any renderer runs. The module docstrings say so, so this is a
known limitation — filing it for the ratio.

A 15-recipe run of ours produces **107 step screenshots and 15 final ones**.
The step sequence is what a reviewer actually opens; it is where the dialog
appears, the selection moves, the error arrives. That is the 107 still
rendering as grey text, while the single end-state image is in colour.

Second-order: step-screenshot dedup fingerprints the rendered grid rather than
the screen text, specifically so a colour-only change counts as a change. Since
step grids are built from already-flattened text, that colour-awareness is
inert for the artifacts dedup actually applies to.

Suggestion: an optional `screen_attributed: AttributedScreen | None` on
`StepResult`, populated by sessions that can supply one — the pty session holds
a live `pyte.Screen`, and the tmux backend has `capture-pane -e`. Renderers
already accept `render_attributed`, so the far end of the pipeline exists;
this is about carrying the grid through the step model.

md-mt:
Fixes #175.

`final.svg` and the `attributed_rsvg` video have rendered from an attributed
cell grid for a while. The per-step images under `steps/` did not, because
`StepResult.screen` is a `str` and a grid rebuilt from flattened text is
monochrome by construction. `StepResult` grows an optional
`screen_attributed: AttributedScreen | None`, filled by sessions that can supply
one, and the per-step screenshot renders from it when it is there.

## Evidence from a real run

Same 13 example recipes, `origin/main` against this branch, same machine:

| | step screenshots | in colour | with text attributes |
|---|---|---|---|
| before | 76 | 0 | 0 |
| after | 76 | **9** | **11** |

Final screenshots are unchanged at 3/13 attributed (1 in colour) — that path
already worked and is untouched. The 9 colour images are every step of
`colour-stress`, up to 400 distinct fills each; the 2 extra attributed ones are
`pi-help` and `pi-list`, which gained `font-weight="700"` on a bold heading. The
other 65 step images are **byte-identical** to the baseline, because those
recipes drive genuinely monochrome TUIs.

## The dedup interaction

The report's second-order point holds, and I confirmed it before fixing it:
`_render_step_screens` fingerprints the rendered grid rather than the text
specifically so a colour-only change counts as a change, and that was inert for
step artifacts because the grid was rebuilt from `StepResult.screen`.

`ColourOnlyChangeEndToEndTest` drives a real pty session painting `STATUS OK` in
green, then repainting the same characters in the same cells in red. All four
steps have byte-identical `.txt`. With `dedup_step_screenshots: true`:

```
before (origin/main)          after
01-green  -> 01-green.svg     01-green  -> 01-green.svg    (#7ee787)
02-flip   -> 01-green.svg     02-flip   -> 01-green.svg
03-red    -> 01-green.svg     03-red    -> 03-red.svg      (#ff7b72)
04-finish -> 01-green.svg     04-finish -> 03-red.svg
1 image written               2 images written
```

Five of the new tests fail if the one-line preference in
`evidence._render_step_screens` is reverted, including that one.

## Which sessions can supply a grid, and which cannot

| session | grid | how |
|---|---|---|
| `pexpect`, `pexpect_asciinema`, `docker` | yes | `TerminalSession.screen_attributed()` off the live `pyte.Screen` |
| `tmux` | yes | `TmuxSession.screen_attributed()` via `capture-pane -e` (already existed) |
| agent-driven mode | yes | one `replay_cast_both` of the operator's cast |
| any third-party `SessionBackend` without `screen_attributed()` | **no** | falls back to the flattened text, monochrome, unchanged |

Two further limits, both pinned by tests rather than only documented:

- **The `png` renderer stays monochrome** whatever the grid contains. It
  implements only the text-only protocol, so a grid cannot reach it. Colour in
  step screenshots is an `svg`-renderer property.
- **Dim (SGR 2) reaches the grid on the tmux path but not the pty one**, because
  pyte 0.8.2's `Char` models no dim field. Unchanged by this, but now it differs
  *between* step screenshots depending on backend, so it is written down.

I have not replaced the existing caveat with a broader claim. `AGENTS.md` now
says to write "when the session supplies a grid" rather than "step screenshots
are in colour", and `test_public_claims.py`'s banned phrasings are untouched and
still pass.

## Memory, measured

A grid per step is not free, and unshared it is not cheap:

| | per 100x32 grid | retained across a 75-step run |
|---|---|---|
| one cell object per cell | 752 KiB | 55 MiB |
| sharing equal cells (this PR) | **47.8 KiB** | **3.5 MiB** |

A terminal screen is mostly repetition, and cells are frozen, so `_CellPool`
gives one object to every cell that compares equal within a grid build. That is
a 15.7x reduction, and it makes the cost scale with the number of *distinct*
cells rather than the number of cells. For comparison the step *text* for that
same run retains 60 KiB. Extrapolated to the reporter's 107 steps: about 5 MiB,
against about 79 MiB unshared. The pool is per grid build and is dropped with
it, so nothing accumulates process-wide. `StepScreenMemoryTest` pins the
property (ratios and orders of magnitude, not byte counts).

## One read, not two

Step capture goes through `screen.capture_screen`, which reads the grid first
and derives the step's text from it with the same normalisation
`screen_text`/`TmuxSession.screen` apply. Two separate reads can describe
different instants — the tmux backend's are two `capture-pane` invocations —
which produces a `steps/NN.txt` describing one screen while `steps/NN.svg` and
the dedup verdict describe another. `ScreenSource::capture_screen` on the Rust
side made the same choice for the same reason. A test pins that
`grid_text(from_pyte(s)) == screen_text(s)` across colour, wide characters,
combining marks and trailing blanks, so `step.screen` has not shifted meaning.

## Parser hardening

`capture-pane -e` returns escape sequences, not a grid, and the parser turning
them into one is now load-bearing for step images. A capture is a snapshot of a
pane still being painted, so a sequence can be cut anywhere. A CSI without its
final byte used to emit its parameter bytes as glyphs — a literal `[31` in the
middle of a screenshot. Fixed, plus a robustness suite over 18 malformed and
partial payloads asserting none raises, none leaks a control byte into a cell,
and each still produces well-formed XML.

## Compatibility

- `screen_attributed` is the **last** field and defaults to `None`, so every
  existing `StepResult(name, passed, detail, screen)` is unmoved.
- `to_dict()` does **not** emit it. `result.json` is a shape shared with the
  Rust implementation and read by the run cache, and nothing downstream of the
  JSON re-renders an image, so a grid there would be bulk with no reader. A
  `StepResult` rebuilt by `from_dict` therefore has no grid — the honest answer
  rather than a lossy one, and the run cache does not need one because its
  artifacts were rendered before it was written.
- `termproof.builtin_steps.step_result(name, passed, detail, session)` is
  exported so a third-party `StepAction` can opt in with a one-line change.

## Rust

**Nothing changes there, checked rather than assumed.** Rust's per-step evidence
has carried a grid since `EvidenceCollector` was written
(`CapturedStep.attributed`), `Session::screen_attributed` is already optional
there with a documented `None` default, and its dedup already fingerprints a
real grid. The serialised `StepResult` shape is untouched by this PR, nothing in
Rust reads Python `StepResult` JSON (`parity.rs` reads only recipe/renderer/
passed/score/assertions; `cache.rs` reads its own entries), and the conformance
corpus records only `name`/`passed`/`detail` — `grep -rn "attributed" conformance/`
is empty.

Worth flagging for a maintainer, not fixed here: the two implementations put the
grid in different places. Python hangs it off `StepResult`; Rust deliberately
keeps it out of `RunResult` in a parallel `CapturedStep`/`EvidenceManifest`
model, with the rationale written out at `collector.rs:49-58`. Nothing forces
convergence today.

## Noted for follow-up, not done here

- **The corpus gate cannot pin a colour step image.** `CorpusByteIdentityTest`
  re-renders each step entry from its recorded `.txt`, which is correct for the
  entries checked in — they predate the grid, and the gate passes untouched —
  but nothing on disk carries a step's grid, so a corpus regenerated from a
  colour-emitting recipe would produce images it cannot reproduce. Recording the
  grid beside the text (an SGR-encoded `.ans` next to each `.txt`) is the fix.
  Written down in the gate's own docstring, in `AGENTS.md` and in
  `docs/evidence-quality.md` rather than left as a trap.
- **A step image and a video frame of the same moment are not the same bytes.**
  `final.svg` and the video both replay the cast; a step grid is read from the
  live session, which is the only place that moment exists. The correspondence
  claim in `cast_video.py` now says so.
- **OSC sequences still leak their body as glyphs** in the ANSI parser.
  `capture-pane -e` emits SGR only, so this is not reachable from the tmux path;
  out of scope here.

## Verification

- `ruff check .`, `mypy termproof`, `unittest discover -s tests` — 718 tests, all
  pass; `scripts/run_stdlib_tests.py` (120) and `scripts/check_version.py` too.
- The two `test_wait_for_regex` / `test_wait_for_idle_step` fakes now set
  `screen_attributed = None` explicitly. A bare `Mock()` auto-creates the
  attribute and answers it with a `Mock`; declaring it makes those fakes model
  what they are, a session with no grid, which is also the fallback path this PR
  most needs covered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,6 +136,16 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   non-letter finals. A fresh `ESC` now also abandons a sequence in progress
   rather than being read as one of its parameter bytes.
 
+- Escape sequences that are not CSI are recognised rather than rendered as
+  glyphs. Only CSI was handled; every other family had its `ESC` dropped and
+  its body printed, so an OSC-8 hyperlink — which `capture-pane -e` really does
+  emit — turned `beforeTXTafter` into `before]8;;url\TXT]8;;\after` in a step
+  screenshot and in the `.txt` beside it. OSC, DCS, SOS, PM and APC are now
+  consumed to either terminator (`BEL` or `ESC \`), as are charset designation,
+  the DEC line-size controls and the two-byte escapes; SS2 and SS3 consume
+  their introducer and leave the character they shift. None of them contributes
+  anything visible, which is what a terminal shows for them too.
+
 ### Python — Changed
 
 - The attributed grid builders share one cell object between cells that compare
@@ -144,12 +154,17 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   retains go from 37.1 MiB to 2.9 MiB — 506 KiB down to 40 KiB per step. That
   is what makes keeping a grid per step affordable.
 
+  The saving is in the repetition and nowhere else, so the floor is exactly no
+  saving: a screen whose 3,200 cells are all distinct measures 527 KiB pooled
+  and 527 KiB unpooled. No real screen looks like that, and the number states
+  the mechanism rather than qualifying it.
+
   Measured with `tracemalloc`, summing allocations made in
   `termproof/attributed.py` that are still alive once the run's results are in
   hand. A `sys.getsizeof` walk of the same objects reads higher — around 46-55
   MiB unshared depending on what the walk counts — because `getsizeof` on a
   key-sharing instance dict reports more than the allocator handed out. Same
-  conclusion, and the ratio is 12-16x on every method tried.
+  conclusion either way.
 
 [#175]: https://github.com/md-mt/termproof/issues/175
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,7 +103,14 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
 
   The field is deliberately absent from `StepResult.to_dict()`: `result.json` is
   a shape shared with the Rust implementation and read by the run cache, and
-  nothing downstream of it re-renders an image.
+  nothing downstream of it re-renders an image. It is `compare=False` for the
+  same reason — equality tracks the serialised shape, so a live result still
+  equals `RunResult.from_dict(result.to_dict())` as it always has.
+
+  `AttributedScreen` also grows a compact `__repr__`. The generated dataclass
+  one was an `AttributedCell(...)` per cell, around half a megabyte for a
+  100x32 grid, and a grid now hangs off every `StepResult` — which is what a
+  failing assertion would have printed.
 
 - `termproof.screen.capture_screen`, one read of a session's screen returning
   text and grid together, and `termproof.builtin_steps.step_result` for step
@@ -122,13 +129,27 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   a literal `[31` in the middle of a screenshot; a terminal waiting for the
   terminator displays nothing, and now neither does the parser.
 
+- A CSI sequence ending in a final byte that is not a letter no longer eats the
+  first letter of the text after it. ECMA-48 puts the final byte at 0x40-0x7E;
+  the parser scanned for `isalpha()`, so `before\x1b[1~after` rendered as
+  `beforefter`. Same for `\x1b[5@`, `\x1b[2^`, `\x1b[?25l` and the rest of the
+  non-letter finals. A fresh `ESC` now also abandons a sequence in progress
+  rather than being read as one of its parameter bytes.
+
 ### Python — Changed
 
 - The attributed grid builders share one cell object between cells that compare
-  equal. A terminal screen is mostly repetition, so this takes a 100x32 grid
-  from roughly 750 KiB to roughly 48 KiB — measured over the example corpus,
-  55 MiB of retained grids for a 75-step run becomes 3.5 MiB — which is what
-  makes keeping a grid per step affordable.
+  equal. A terminal screen is mostly repetition, so a typical 100x32 grid goes
+  from 527 KiB to 31 KiB, and the 75 step grids a run over the example corpus
+  retains go from 37.1 MiB to 2.9 MiB — 506 KiB down to 40 KiB per step. That
+  is what makes keeping a grid per step affordable.
+
+  Measured with `tracemalloc`, summing allocations made in
+  `termproof/attributed.py` that are still alive once the run's results are in
+  hand. A `sys.getsizeof` walk of the same objects reads higher — around 46-55
+  MiB unshared depending on what the walk counts — because `getsizeof` on a
+  key-sharing instance dict reports more than the allocator handed out. Same
+  conclusion, and the ratio is 12-16x on every method tried.
 
 [#175]: https://github.com/md-mt/termproof/issues/175
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,57 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   termproof`.** It asserted the command's absence because the package was
   unpublished, and went on passing after the publish made that wrong.
 
+### Python — Added
+
+- `StepResult.screen_attributed`, an optional `AttributedScreen | None`
+  alongside the existing flattened `screen`. When it is present the per-step
+  screenshot is rendered from it, so the images under `steps/` carry the colour
+  and text attributes `final.svg` already did ([#175]).
+
+  Every built-in session backend fills it — `pexpect`, `pexpect_asciinema` and
+  `docker` from the session's `pyte.Screen`, `tmux` from `capture-pane -e` —
+  and the agent-driven mode fills it from its cast replay. Over the 13 example
+  recipes, step screenshots carrying attributed rendering went from 0/76 to
+  11/76 (9 in colour, 11 with text attributes); the other 65 are byte-identical
+  because those recipes drive genuinely monochrome TUIs.
+
+  Optional means optional. A `SessionBackend` whose session does not implement
+  `screen_attributed()` reports no grid, and its step screenshots render from
+  the text exactly as before. The `png` renderer takes text only, so PNG step
+  screenshots stay monochrome either way, and dim (SGR 2) reaches the grid on
+  the tmux path but not the pty one.
+
+  The field is deliberately absent from `StepResult.to_dict()`: `result.json` is
+  a shape shared with the Rust implementation and read by the run cache, and
+  nothing downstream of it re-renders an image.
+
+- `termproof.screen.capture_screen`, one read of a session's screen returning
+  text and grid together, and `termproof.builtin_steps.step_result` for step
+  actions that want the same. The grid is read first and the text derived from
+  it, so `steps/NN.txt` and `steps/NN.svg` cannot describe different instants.
+
+### Python — Fixed
+
+- Step-screenshot dedup fingerprints the rendered grid so that a colour-only
+  change between two steps counts as a change. That was inert for the artifacts
+  dedup applies to, because the grid was rebuilt from already-flattened text; it
+  now compares the grid the session reported ([#175]).
+
+- A CSI escape sequence cut before its final byte no longer emits its parameter
+  bytes as glyphs. `\x1b[31` at the end of a `capture-pane -e` line rendered as
+  a literal `[31` in the middle of a screenshot; a terminal waiting for the
+  terminator displays nothing, and now neither does the parser.
+
+### Python — Changed
+
+- The attributed grid builders share one cell object between cells that compare
+  equal. A terminal screen is mostly repetition, so this takes a 100x32 grid
+  from roughly 750 KiB to roughly 48 KiB — measured over the example corpus,
+  55 MiB of retained grids for a 75-step run becomes 3.5 MiB — which is what
+  makes keeping a grid per step affordable.
+
+[#175]: https://github.com/md-mt/termproof/issues/175
+
 ## [0.3.4] — 2026-08-17
 
 One project, one version. The two repositories became one: the Rust
--- a/README.md
+++ b/README.md
@@ -95,9 +95,14 @@ The Python implementation writes, per run:
 
 Two things worth knowing before you promise them to anyone:
 
-- **The per-step screenshots under `steps/` are monochrome.** They are rendered
-  from plain text. `final.svg` and the `attributed_rsvg` video backend render
-  from the attributed grid and carry colour; the per-step images do not.
+- **Per-step screenshots carry colour only when the session can report a
+  grid.** All four built-in session backends can — pty, pty+asciinema, Docker
+  and tmux — and so does the agent-driven mode, so on a stock install the
+  `steps/` images are as attributed as `final.svg`. A third-party session
+  backend that does not implement `screen_attributed()` still works and still
+  gets its screenshots, in monochrome. Two limits hold either way: the `png`
+  screen renderer takes text only, so PNG step images are always monochrome,
+  and dim (SGR 2) survives on the tmux path but not on the pty one.
 - **Video needs external tools.** `agg` + `ffmpeg` for the default backend, or
   `rsvg-convert` + `ffmpeg` for `attributed_rsvg`. A prebuilt `agg` is bundled
   in the platform wheels.
--- a/python/AGENTS.md
+++ b/python/AGENTS.md
@@ -75,15 +75,33 @@ not a tracked file, so it stays a manual check.
 Two limits, both pinned by tests; check them before making any blanket claim
 about colour in the artifacts:
 
-- **Only `final.svg` and the `attributed_rsvg` video render from the grid.** Step
-  screenshots render from `StepResult.screen`, which is pyte's flattened
-  `display`, so they are monochrome — and they are the large majority of the
-  rendered corpus. The dedup tests that inject SGR escapes into a `StepResult`
-  are helper coverage of `_render_step_screens`, not evidence that a real step
-  screenshot has colour.
-- **Dim (SGR 2) is lost on the cast-replay path**, because pyte 0.8.2's `Char`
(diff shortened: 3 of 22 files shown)


md-mt:
Round 1 addressed at `6f6e460`, rebased onto `main` (#176). Both findings fixed, memory figures re-measured with a stated method. No change to what the PR does.

### 1. `screen_attributed` is now `compare=False`

Confirming the intent you asked about: yes. The grid is evidence a step carries, not part of its identity — two steps with the same verdict, detail and screen text are the same result. Equality should track the shape callers can see, which is the serialised one. Bonus: `__hash__` no longer walks every cell.

Your four probes at this head:

```
plain_eq_from_dict            True
grid_eq_from_dict             True
plain_eq_grid                 True
run_with_grid_eq_roundtrip    True
```

Pinned by `test_equality_tracks_the_serialised_shape` and `test_a_whole_run_result_equals_its_round_trip`.

**One thing `compare=False` does not fix, so I fixed it too.** The generated dataclass repr of an `AttributedScreen` is one `AttributedCell(...)` per cell — **490 KB for a single 100x32 `StepResult`**. With `compare=False` an equality failure is now about the *other* fields, but unittest would still print both reprs. `AttributedScreen` gets a compact `__repr__` (dimensions, cursor, first line); `rows` and `text_lines()` remain for when the cells are the question. `StepResult` keeps the field in its repr, so you can still see whether a grid is attached.

### 2. CSI final bytes are now the real range

`_consume_ansi_sequence` ends at 0x40–0x7E rather than `isalpha()`. Your three payloads:

```
'before\x1b[1~after'  =>  'beforeafter'
'before\x1b[5@after'  =>  'beforeafter'
'before\x1b[2^after'  =>  'beforeafter'
```

`isalpha()` was also true of non-ASCII letters, which cannot terminate a CSI, so the range check is narrower and wider in the right directions.

**An edge the naive range fix introduces, handled:** `[` is 0x5B, inside the final-byte range. So `\x1b[31\x1b[32mX` would have ended its first sequence on the second `[` and emitted `32mX` as text — a regression against the old behaviour. A fresh `ESC` now abandons the sequence in progress and is re-read from its own start; that case renders `X` with `fg=green`.

Tests: the three payloads plus `\x1b[H`, `\x1b[?25l`, `\x1b[1;2H`; a sweep asserting **every** byte in 0x40–0x7E terminates; a check that parameter bytes (0x30–0x3F) do not; the abandoned-sequence case; and seven new entries in the malformed-payload suite covering non-letter finals, intermediates and abandoned sequences, with a new assertion that the text either side of an escape survives intact.

### 3. Memory figures re-measured, method stated

You measured 3.24 MiB against 45.88 MiB; I had 3.5 against 55. Both are `sys.getsizeof` walks, and `getsizeof` **overstates** — on a key-sharing instance dict it reports more than the allocator actually handed out, and counting `cell.text` per cell double-counts shared strings. That is the whole delta between our numbers.

`tracemalloc` is the less arguable measurement: allocations made in `termproof/attributed.py` that are still alive once the run's results are in hand. That is what the changelog now quotes, with the method written beside it:

| | unshared | pooled | ratio |
|---|---|---|---|
| 75 step grids, example corpus | 37.1 MiB | **2.9 MiB** | 12.6× |
| per step | 506 KiB | **40 KiB** | |
| one typical 100×32 grid | 527 KiB | **31 KiB** | 17.1× |
| one 100×32 grid, every cell distinct | 527 KiB | 292 KiB | 1.8× |

That last row is the honest floor: the saving is in repetition, not in the cells. Every method any of us tried lands between 12× and 16×, so the conclusion never moved — but you are right that a number in a PR body is a claim, and the previous ones were a method I had not named.

`StepScreenMemoryTest` keeps its own `getsizeof` bound and now says why it reads higher than the changelog. A machine-dependent byte count has no business failing a suite, so the test asserts the property (cost scales with *distinct* cells) rather than the figures.

### Left as you accepted it

The corpus-regeneration trap stays a documented follow-up, unexpanded.

### Re-verified at this head

`ruff`, `mypy`, **728 tests**, the stdlib-only subset (125) and `check_version.py` all pass. The headline result is unchanged by the parser fix: 0/76 → 9/76 step screenshots in colour, 0/76 → 11/76 attributed, finals still 3/13. No files under `rust/` touched.

md-mt:
Round 2 addressed at `05e567e`. The blocking finding was right, and it was worse than a gap — it was a regression I introduced, exactly as you characterised it.

## Blocking: non-CSI escapes

You asked for the sibling and I did not find it. Reproduced before fixing, and it was **not only OSC-8** — every non-CSI family leaked its body:

```
OSC-8 hyperlink  'before]8;;http://x\TXT]8;;\after'
OSC title (BEL)  'before]0;titleafter'
DCS              'beforeP1$r0m\after'
charset ESC ( B  'before(Bafter'
ESC 7 / ESC # 8  'before7after' / 'before#8after'
APC / PM         'before_apc\after' / 'before^pm\after'
RIS ESC c        'beforecafter'
```

Each family now gets a decision rather than a default:

| family | handling | rationale |
|---|---|---|
| OSC, DCS, SOS, PM, APC | consume to terminator — **both** `BEL` and `ESC \` | body is a URL, title or device reply; nothing visible. Both terminators are in real use |
| charset designation, DEC line size, two-byte escapes | consume intermediates `0x20–0x2F` then final `0x30–0x7E` | ECMA-48 gives them one shape, so one branch |
| SS2 / SS3 | consume the **introducer only** | the character it shifts *is* displayed — the one family whose next byte survives |
| cut before terminator | consume rest of line | same rule the truncated CSI already used |
| fresh `ESC` mid-sequence | abandon, re-read from the new ESC | same rule CSI already used |

Through the public tmux backend, your exact reproducer:

```
session.screen           : 'beforeTXTafter'
step_result(...).screen  : 'beforeTXTafter'      <- was 'before]8;;http://…\TXT]8;;\after'
EQUAL to plain capture   : True
colour still reaches grid: ['red', 'red', 'red']
```

## The root you pointed at, audited

> satisfy yourself that no other path in this change replaces a known-good plain-text source with a grid-derived one

There is **exactly one** such substitution — `capture_screen`, `screen.py:111` — and its cost differs by backend, which is the part I had not written down. `capture_screen`'s docstring now says so:

- **pty** — the grid is the same `pyte.Screen` that `screen_text` flattens. No parser involved; the two cannot disagree. Already pinned.
- **tmux** — the grid is parsed from `capture-pane -e` while `session.screen` is `capture-pane` already flattened *by tmux*. This is a claim about a parser, and the claim was false. **This is the one that broke.**
- **agent-driven** — takes its text from `screen_text`, not from its grid. Not a third case.

So the guard is `TmuxTextEquivalenceTest`: it compares the two readings **against each other** — not against a literal — through the public backend, over eight payloads covering every escape family. Against the round-2 parser it fails on the hyperlink payloads exactly as you reported:

```
FAIL test_a_hyperlink_does_not_put_its_url_on_the_screen
FAIL test_the_derived_text_matches_the_plain_capture (payload='osc 8 hyperlink')
FAIL test_the_derived_text_matches_the_plain_capture (payload='mixed')
```

A future gap in the parser now fails a test rather than appearing in evidence.

Also added: 18 per-family unit cases, SS2/SS3, string sequences cut before their terminator, abandonment, a check that a string sequence does not disturb surrounding SGR, and 19 new fuzz payloads with an assertion that no OSC body (`http://`, `window title`, `id=x`, …) reaches a cell.

(Three of my own new fuzz payloads were fixture bugs, now fixed: two are deliberately truncated and correctly render `before` with nothing after, and one had a literal space I had not accounted for.)

## Non-blocking: the memory floor

You are right and the correction is a better argument than the claim. My 1.8× came from a screen whose cells were **not** in fact all distinct — random 256-colour runs over repeated 10-character groups, so plenty of cells still compared equal. With 3,200 genuinely unique cells, by the stated `tracemalloc` method:

| | unshared | pooled | ratio |
|---|---|---|---|
| 75 step grids, example corpus | 37.06 MiB | **2.94 MiB** | 12.6× |
| per step | 506 KiB | **40 KiB** | |
| one typical 100×32 grid | 526.8 KiB | **30.8 KiB** | 17.1× |
| one 100×32 grid, every cell distinct | 526.6 KiB | 526.6 KiB | **1.0×** |

Corrected in the changelog and in the `_CellPool` prose. As you say, 1.0× states the mechanism exactly: the pool shares equal cells, so with nothing equal there is nothing to share.

Your full-run figures (37.29 / 3.18 MiB) and mine (37.06 / 2.94) agree to machine variance on the same method.

## Re-verified at this head

`ruff`, `mypy`, **737 tests**, stdlib-only subset (131), `check_version.py` — all pass. Headline unchanged: 0/76 → 9/76 step screenshots in colour, 0/76 → 11/76 attributed, finals 3/13. No files under `rust/` touched. Corpus-regeneration trap still a documented follow-up, unexpanded.

md-mt:
CI on `05e567e` is green after a re-run. The three red jobs were an infrastructure stall, not the change — recording the diagnosis since "three jobs failed" reads like a code problem.

All three (py3.12, py3.13, evidence) died at the **same step**, `sudo apt-get update && sudo apt-get install -y ffmpeg`, with the Ubuntu mirror going quiet mid-fetch:

```
py3.12   apt step started 07:09:04, last output 07:09:42, job killed 13:09:27
py3.13   apt step started 07:09:06, last output 07:09:44, job killed 13:09:27
verify   apt step started 07:09:52, last output 07:10:31, job killed 13:10:05
```

Six hours is the GitHub job ceiling, so these were killed, not failed — the workflow run's conclusion is `cancelled`. No test output appears in any of the three logs: the suite never started. py3.11 cleared apt before the mirror stalled and passed in four minutes, on the same commit and the same image.

The apt step runs before `uv sync`, so no code from this branch executed in any of the three. Confirmed independently by running the full suite locally on the two interpreters that were killed: **737 tests pass on 3.12 and on 3.13**.

Re-ran the workflow, nothing else changed. 12 passed, 2 skipped, 0 failed.

Unrelated observation, deliberately not fixed here: that apt step has no timeout, so a mirror stall costs a six-hour job slot rather than failing fast. `timeout-minutes` on the job, or a retry around the install, would bound it. Out of scope for this PR.

md-mt:
## Review summary

Three rounds.

### Round 1 — merge-with-follow-ups (0 blocking, 2 non-blocking)

The colour claims were reproduced independently: both trees run on the same machine gave 76 step
screenshots before and after, **0 → 9 in colour**, 0 → 11 with text attributes, and the other 65
byte-identical because those recipes drive genuinely monochrome sources. Frozen cells confirmed
non-mutable and non-aliasing; the dedup fix confirmed in both directions; the optional-session
fallback confirmed with a constructed backend; 108 fuzz payloads produced no crash, hang, control-byte
leak or malformed XML.

Two findings:

1. `screen_attributed` was excluded from `to_dict()` but still participated in dataclass equality, so
   a live result did not equal its own serialised round trip.
2. The parser scanned to `isalpha()` for a CSI terminator, so a sequence ending in a non-letter
   consumed the following character: `before\x1b[1~after` rendered as `beforefter`.

### Round 2 — do-not-merge (1 blocking, 1 non-blocking)

Both round-1 items were fixed. Widening the terminator to `0x40–0x7E` introduced a hazard the author
caught themselves — `[` is `0x5B`, inside that range, so two adjacent sequences would have terminated
the first early; a fresh `ESC` now abandons the sequence in progress.

The blocking finding was the sibling: `_consume_ansi_sequence` handled **CSI only**. `capture-pane -e`
emits non-CSI sequences, OSC-8 hyperlinks among them, and those rendered as visible glyphs. Because
this change derives `StepResult.screen` from the grid, the corruption reached both the step text and
the screenshot — a regression against `main`, where the same input was correct:

```
program:                  printf 'before\033]8;;url\033\\TXT\033]8;;\033\\after\n'
session.screen:           'beforeTXTafter'
step_result(...).screen:  'before]8;;url\TXT]8;;\after'
```

Also: the all-distinct memory floor was reported as 1.8×; measured by the stated method it is 1.0×,
which is the expected property — the pool shares equal cells, so with none equal there is nothing to
share.

### Round 3 — merge (0 blocking, 0 non-blocking)

Verified through the real tmux backend rather than by reading the parser: `session.screen` compared
against `step_result(...).screen` across **93 inputs**, no mismatches. Coverage included OSC with both
terminators, unterminated OSC bodies to end of input, DCS/SOS/PM/APC, SS2/SS3, charset designations,
DEC line-size and bare two-byte escapes, `ESC` as the final byte, `ESC` followed by `ESC`, sequences
split across line boundaries, and the entire round-2 CSI surface re-run.

Root cause containment confirmed: `capture_screen` is the only known-good-text-to-grid-derived-text
substitution; the agent-driven path still sets `StepResult.screen` from the text and carries the grid
separately.

The memory floor now reads 1.0× in both the changelog and the `_CellPool` prose. The headline counts
are unmoved. The full suite passes on 3.12 and 3.13, 737 tests each.

### Note on CI

Three jobs on this branch were killed at the six-hour ceiling while stalled on the system-dependency
install — run conclusion `cancelled`, no test output, suite never started. Not a code failure; the
same commit passed on re-run with no change. Tracked in #183.

### Deferred

The conformance corpus cannot pin a colour step screenshot, so regenerating it from a colour recipe
would break the byte-identity gate. Documented in the test, `python/AGENTS.md` and
`docs/evidence-quality.md` rather than left silent; the failure would be loud at regeneration time.

PR #189, opened 2026-08-19
api: additive seams for an imperative consumer — capture_text, recordings, Selectable, Comparable

md-mt:
Additive API for a consumer whose recipes are **imperative** — they branch on what the screen shows, so they cannot be expressed as the declarative `Recipe` this crate models. Today such a consumer maintains a parallel copy of the evidence, selection and before/after layers; these are the seams that let it use the library instead.

**Goes into 0.4.0**, which main already carries in both manifests and has not yet tagged. Based on `main`; nothing here bumps a version.

---

## The Rust half was written without a toolchain

The branch this came from was authored on a host where `cargo` could not be installed: `static.rust-lang.org` and `sh.rustup.rs` were both unreachable. The Rust code was compiled — by copying the sources over a vendored checkout and building that — but **no Rust test, lint, format check or semver check on it had ever run.** The `#[cfg(test)]` blocks were unexecuted code. Python was fully tested (757 tests, ruff and mypy clean) and reproduced identically here.

Everything below is what running those gates found. Reviewers should not have to infer which parts were verified.

### What a compile-only check missed

| Gate | Result |
|---|---|
| `cargo fmt --all --check` | **1 violation**, in the never-run test code. Fixed. |
| `cargo test --workspace` | **All 9 new tests passed first run.** No fix needed. |
| `cargo clippy --all-targets --all-features -D warnings` | Clean. The `#[allow(dead_code)]` the author was unsure about is correct. |
| 16-combination feature powerset | Green on the branch as written. *I* broke 8 of them by adding an integration test that names `evidence`; gated and fixed in the same commit. |
| Dependency-floor run | Green. |
| `cargo deny` | Clean. |
| **`cargo semver-checks`** | **Failed.** One major-version finding — see below. |

The honest summary is that the unexecuted test code was *nearly* fine: one formatting nit, and no logic error. What the compile could not see was the API compatibility question.

### The finding: this is a breaking change

`cargo semver-checks` reports `function_requires_different_generic_type_params` against `select`, `select_names`, `compute_deltas` and `build_before_after`, and asks for a major bump. A control run on the then-current `main` (`a729c71`, 0.3.4) was clean at 196/196, so the finding is this branch's and not inherited.

I did not take the tool's word for it, because a green semver check is [not trustworthy](https://github.com/obi1kenobi/cargo-semver-checks) as a sole gate. Compiling real consumers against both `main` and this branch:

- **`BeforeAfterResult<R = RunResult>` is fine.** A consumer naming the bare type in a struct field, a return position, an argument, a `let` annotation, a struct literal and nested inside another generic compiles unchanged. The default type parameter does its job.
- **The four functions are not.** These compile on `main` and fail here:
  ```rust
  compute_deltas(&[], &[])
  build_before_after(vec![], vec![])
  select_names(&[], &[], cfg)
  ```
  `error[E0283]: type annotations needed`. The element type used to be fixed by the signature and is now inferred; an untyped empty literal gives inference nothing to work with. A consumer with such a call annotates the slice. There are none in this repository.

Rust has no default type parameter for a *function* — that is an unstable feature — so no arrangement of defaults avoids this. The alternative is keeping the concrete functions and adding generic twins under new names, which is a worse API than the one the commit argues for at length.

### Why this rides on 0.4.0

Because of the above, this cannot land as a patch. #186 has merged and cut **0.4.0** as a breaking release for the dependency-wrapping work; 0.4.0 is not yet tagged, so this ships in it. That means one breaking release instead of two, and it is why `cargo semver-checks` is green: at 0.3.4 → 0.4.0 it reports *major change* and skips all 254 checks.

**That green is worth reading carefully.** It does not mean "nothing broke" — it means "the version already accounts for a break". The evidence that the break is real and bounded is the consumer compile above, not the check.

I did not touch the version. It is already 0.4.0 on `main`.

---

## What is in it

| Commit | What |
|---|---|
| `9844269` | `EvidenceCollector::capture_text` — record a screen the caller already holds |
| `127eabc` | `Recording`, `attach_recording`, `recordings` on the manifest |
| `fed1953` | `selection::Selectable`, `before_after::Comparable` — generic over the caller's types |
| `548a845` | `python/termproof/collector.py` |
| `65b8e08` | Python `attach_to`, and which of `collector`/`evidence` is for what |
| `459392f` | The cross-implementation manifest harness, and the divergence it found |
| `cb41c43` | Changelog, merged into `## [0.4.0]` |
| `b3d2451` | Pin `Recording`'s derived `Debug`, and the four-door audit |

Each commit message carries its own reasoning; `fed1953`'s was amended with the semver finding above.

Changelog entries are merged into `## [0.4.0]`'s existing per-implementation sections rather than appended as new ones — the file's header says a heading is split by implementation, and a second `### Rust — Added` under one version is two lists nobody reads as one. The breaking generics entry sits under `### Rust — Changed (breaking)` beside the dependency wrapping, which is where a reader of the 0.4.0 notes needs to find it.

### Dropped from the original branch

The branch also narrowed `fancy-regex` to `>=0.16, <0.17`. **Dropped.** #179 reaches the same constraint (`^0.16`) from the same evidence, independently, and was stacked on #186 — which existed because narrowing that requirement is a breaking change on its own while a public function returns the dependency's type. Landing the narrowing here, without the wrapping, would have reintroduced the defect two PRs were spent removing. The reasoning is preserved in #179; nothing is lost by dropping it here.

The scratch handoff note the branch carried is deleted rather than landed.

---

## No new public API names a third-party type

Re-audited against 0.4.0's stricter policy, which names **four** doors a dependency can reach the public API through — return type, argument, re-export, and trait impl — and removed the `pub use fancy_regex` / `jsonschema` / `vt100` re-exports I had originally cited as precedent. Done with rustdoc JSON rather than by eye:

- **Return type and argument.** No non-`std`, non-`termproof` type is reachable from any signature this branch adds. `Selectable` and `Comparable` have no associated types and no bounds naming a dependency; their accessors return `&str`, `&[String]` and `bool`. `Recording`'s five fields are all `String`/`Option<String>`.
- **Re-export.** None added. 0.4.0 removed three, and nothing here reinstates them.
- **Trait impl.** `Recording`'s *declared* trait set — `Debug`, `Clone`, `PartialEq`, `Eq`, `Serialize`, `Deserialize` — is byte-identical to `PublishedStep`'s, which predates this branch and sits in the same manifest. serde is the sanctioned exception (1.x, so cargo unifies every consumer onto one copy). `BeforeAfterResult` declares only `Debug` and `Clone`.

One caveat worth stating because it looks alarming in raw rustdoc output: `Recording` renders blanket impls from `zerocopy`, `tracing`, `tower_http`, `downcast_rs`, `typenum`, `dyn_clone`, `hashbrown` and `indexmap`. Those are blanket impls those crates provide for every `T`, not something this branch declares — the list is byte-identical on `PublishedStep`, so it is a property of the dependency graph and not a door anything here opens.

`Recording` derives `Debug`, and 0.4.0's fourth door is exactly that: `PyRegex`'s derive was printing a pattern whose rendering moves across `fancy-regex` *patch* releases. The derive is safe here only because every field is a `String` — which was an argument rather than a check, so `8271e82` pins the exact `Debug` string the way #186 pinned `PyRegex`'s. Adding a field of any other type now fails a test instead of silently widening the surface.

`ScreenCapture`, `CaptureKind` and `RawOutput` are unchanged by this branch; `ScreenCapture.attributed` is `Option<AttributedScreen>`, this crate's own type.

## The cross-implementation manifest diff

`CrossImplementationShapeTest` asserted the manifest key set matched Rust's *by spelling the keys out* — a list derived by reading the Rust structs, never by running them. With a toolchain available, I generated a manifest from each implementation against the same input and diffed.

**The manifests agree byte-for-byte.** Every key, every value, the `step-NN-label` filename scheme, the dedup verdict and its `same_as` back-reference, the `kind` spelling, both recordings, and the omission of `recordings` when empty. The hand-written list was right.

**The files they point at did not.** Python appended a trailing newline to every `step-NN.txt`; Rust writes the screen verbatim. Four files differed underneath manifests that agreed exactly.

Python is the side that changed, per the rule that Python is wrong when the two disagree — and on the merits: the screen is what an assertion was evaluated against, and Rust records it unaltered. Only reachable through `termproof.collector`, which is new in the same release, so no published artifact changes shape.

The check now lives in `conformance/` as the third differential harness, same two-half shape as the step and assertion ones, and it compares **the written file contents as well as the document** — that is what caught the newline, and a manifest is a set of paths, so agreeing on paths is not agreeing on files. Two machine-dependent things are neutralised identically on both sides (the `rsvg-convert` shell-out, and the temporary publish directory). Verified to fail before being trusted: perturbing one label in the recorded document turns it red.

`EVIDENCE_MANIFEST_VERSION` does not move. Both tests that pin the empty-`recordings` omission pass.

---

## The Python collector overlap

`collector.py` is new and unused, and `termproof.evidence` still does an overlapping job with its own step-screenshot dedup sharing no code. **Documented rather than consolidated**, and the docstring says why rather than leaving it implied.

The two are not the same job. A table in `collector.py` now sets out what each is driven by, what it writes, how it fingerprints for dedup and what it records, what a render failure does to each, and which Rust module each corresponds to — they differ in every row. I read #181's change to `evidence`'s dedup first, as instructed; it is further from the collector's than the branch assumed (`evidence` parses ANSI to build its grid and is config-gated off by default; the collector's is unconditional and uses the source's own grid).

Folding `evidence` onto `collector` would change the file layout `render_artifacts` has always written — `steps/`, `steps-manifest.json`, `unchanged_from_previous` — which every existing reader of a run directory depends on. That is a separate change with its own compatibility question, not a tidy-up to fold into this one.

**Python's `EvidenceManifest` gained `attach_to`.** Python's `RunResult` carries the same `recipe_name`/`renderer`/`artifacts` contract Rust's `RunIdentity::matches` compares, so it ports exactly, raising `ValueError` where Rust returns `Err`. Five tests, including that another run's evidence is refused and that the result's existing artifacts survive. `RunIdentity`'s docstring already promised the method existed.

---

## Not done, deliberately

- **`ScreenSource` is not widened to three methods.** Three getters against a live program describe three different instants. The consumer adapts.
- **`Recipe` stays declarative.** No convergence of the recipe model.
- **`EVIDENCE_MANIFEST_VERSION` is not bumped.**
- Nothing merged, tagged or published.

## Verification

Every gate, on the final stack:

```
cargo fmt --all --check                                          clean
cargo test --workspace                                           555 passed, 0 failed
cargo clippy --workspace --all-targets --all-features -Dwarnings  clean
16-combination feature powerset (clippy + test each)             16/16 green
dependency-floor run (fancy-regex =0.16.2, jsonschema =0.32.0)   547 passed
cargo deny check                                                 advisories/bans/licenses/sources ok
cargo semver-checks check-release -p termproof                   no update required (major change)
uv run coverage run -m unittest discover -s tests                779 passed
uv run ruff check . && uv run mypy termproof                     clean
scripts/check_schema_copies.py && scripts/check_version.py       clean
```

Python matched the authoring machine exactly (757 → 762 with the new tests), so no environmental divergence to report.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -137,6 +137,29 @@ between releases, and `(?<=a+)b` is rejected at 0.16 and accepted at 0.19 —
 which is why the requirement is pinned to one minor and the differential
 harnesses are run against it rather than assumed.
 
+- **`select`, `select_names`, `compute_deltas` and `build_before_after` are
+  generic.** Each took a concrete slice and read two or three fields of it;
+  each now takes `R: Selectable` or `R: Comparable`. Every call that names its
+  element type compiles unchanged and resolves to the same types it did before,
+  and `BeforeAfterResult` on its own still means `BeforeAfterResult<RunResult>`
+  — verified by compiling a consumer that names the bare type in a struct
+  field, a return position, an argument, a `let` annotation, a struct literal
+  and nested inside another generic.
+
+  **What breaks:** a call that passed an untyped empty literal, as in
+  `compute_deltas(&[], &[])` or `select_names(&[], &[], &cfg)`. The element
+  type used to be fixed by the signature and is now inferred, and an empty
+  literal gives inference nothing to work with. Annotate the slice —
+  `compute_deltas(&[] as &[RunResult], …)` — or pass a typed binding. There are
+  no such calls in this repository.
+
+  Rust has no default type parameter for a function, so no arrangement of
+  defaults avoids this; `cargo semver-checks` reports it as
+  `function_requires_different_generic_type_params` and asks for a major bump,
+  which under this project's pre-1.0 rule is the minor digit. It therefore
+  belongs in the same breaking release as the wrapping above rather than asking
+  for a second one.
+
 ### Rust — Docs
 
 - **The crate docs used to claim `schemars` was the only dependency reaching
@@ -231,6 +254,13 @@ harnesses are run against it rather than assumed.
   the schema too. `load_recipe_schema()` lost its `docs/` fallback, which
   existed only because the resource was conditional (#174).
 
+- **A published step's `step-NN.txt` no longer gains a trailing newline.** The
+  screen is written verbatim, as the Rust implementation writes it. Found by
+  running both implementations over one scenario and diffing what they wrote:
+  the manifests agreed byte-for-byte and four of the files they pointed at did
+  not. Only reachable through `termproof.collector`, which is new in the same
+  release, so no published artifact changes shape.
+
 ### Rust — Fixed
 
 - `schema::load_canonical_schema` returns the canonical schema for every
@@ -250,6 +280,45 @@ harnesses are run against it rather than assumed.
   in 0.3.4 because it read outside the crate; it no longer does, so the seam is
   testable from the artifact that is published — including the decoy-directory
   regression (#174).
+
+- **`EvidenceCollector::capture_text`** records a screen the caller already
+  holds. `capture` and `capture_failure` pull from a `ScreenSource`, which
+  presumes something live to pull from; text recovered from a log, the screen a
+  step returned before the session moved on, or a golden file in a test arrives
+  no other way, and recording one meant writing a throwaway `ScreenSource`
+  whose only job was to hand back a string. It is a step like any other — same
+  sequence, same index, same filename scheme — so the manifest does not develop
+  a gap where one was taken. No raw output log is attached even for
+  `CaptureKind::Failure`: there is no source to ask for one.
+
+- **`Recording`, `EvidenceCollector::attach_recording` and
+  `EvidenceCollector::recordings`** carry whole-session recordings into the
+  manifest. A collector's captures are instants; a run also produces a span —
+  the terminal recording and whatever video was encoded from it — and a caller
+  keeping both had to write a second document beside the manifest and join them
+  by convention. The collector does not *produce* recordings: encoding a cast
+  is a tool-shelling job with its own failure modes and its own choice of tool.
+  One `error` field rather than separate conversion and upload errors, because
+  a conversion that failed leaves nothing to upload. `recordings` is
+  `skip_serializing_if = "Vec::is_empty"`, so a run that records nothing writes
+  byte-for-byte the document it wrote before — which is why
+  `EVIDENCE_MANIFEST_VERSION` does not move, and it is asserted by a test.
+
+- **`selection::Selectable` and `before_after::Comparable`** let both modules
+  work on results that are not this crate's. A suite whose recipes branch on
+  what the screen shows cannot be expressed as a declarative `Recipe`, but it
+  has names and `ci_paths` and still has the problem of running everything on
+  every change; a caller whose results carry fields this crate does not model
+  had to convert to a `RunResult` and back to ask whether anything flipped,
+  losing those fields on the way. One accessor per field actually read.
+  `(String, Vec<String>)` implements `Selectable` too, so a caller can ask the
+  question without writing an impl — that pairing is what the Python
+  implementation's `select_names` has always taken.
+
+  No third-party type appears in any of this. `Recording` holds `String` and
+  `Option<String>` and derives nothing that reaches past them; `Selectable` and
+  `Comparable` are generic over the caller's own types and their accessors
+  return `&str`, `&[String]` and `bool`.
 
 ### Changed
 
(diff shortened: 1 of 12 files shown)


md-mt:
## Review summary

Verdict: **merge** — 0 blocking, 0 non-blocking. Reviewed at `aa12784b`.

This PR was reviewed once, thoroughly, rather than iteratively. It is the largest change in the 0.4.0
release (+2099/−23 against `main` at review time) and the only one landing new public API immediately
after a release whose purpose was *removing* public API. That framing drove the review.

### The public API audit — the point of the exercise

Every new or changed public item was checked for whether it names, accepts, returns or stores a
third-party type:

| Item | Third-party exposure |
|---|---|
| `EvidenceCollector::capture_text` | none |
| `Recording` and its builders | serde derives only |
| `EvidenceCollector::attach_recording` / `recordings` | none |
| `EvidenceManifest::recordings` | none beyond serde |
| `selection::Selectable` and its impls | none |
| `select` / `select_names` | no third-party bounds |
| `before_after::Comparable` and its impls | none |
| `BeforeAfterResult<R = RunResult>` | no bound or type leak |
| `compute_deltas` / `build_before_after` | no third-party bounds |

**Nothing leaks.** The single third-party trait exposure is `serde` on `Recording`, which matches the
existing manifest structs and the crate's documented exception.

`Recording` derives `Debug`, and a derived `Debug` reaching an inner dependency was a real defect
found earlier in this release. Here it is pinned by
`a_recordings_debug_reaches_nothing_but_its_own_strings`.

### The paper-over audit

The Rust half of this branch was authored on a machine without a toolchain, so **no Rust test on it
had ever run** until shortly before review. That makes weakened assertions the specific risk, and the
review searched for them: loosened expectations, `#[ignore]`, broad new allows, skipped tests,
expectations rewritten to match observed output.

None found. The only new Rust allow is `#[allow(dead_code)]` on a test-only field, present to prove
`Comparable` preserves fields belonging to a foreign result type. The new broad exception handlers on
the Python side are in best-effort render and upload paths, documented and tested as part of the
collector contract rather than added to quiet a failure.

### The rebase

The branch rebased mid-review, from `b3d2451` to `aa12784b`.
`git range-diff` reports all eight commits patch-equivalent; the only delta is inherited base changes
from #190 and #179. The PR's own diff against `main` is unchanged.

### Scope discipline

`EVIDENCE_MANIFEST_VERSION` is unmoved — `recordings` is omitted from the JSON when empty on both
sides specifically so it does not have to change, and two tests pin that. `ScreenSource` was not
widened to three getters; three getters against a live program describe three different instants, and
the result is a manifest that validates and lies. The recipe model stays declarative.

### Where this lands

Into 0.4.0, which is written but not yet tagged. Changelog entries sit under the existing `[0.4.0]`
heading rather than `[Unreleased]`, and no version was bumped.

PR #190, opened 2026-08-19
ci: bound every job, and take the package mirror off the critical path (#183)

### LINKED ISSUE #183: Python CI stalls on the system-dependency install and burns the six-hour job ceiling
Python CI jobs intermittently stall on the system-dependency install step and
are killed by the six-hour job ceiling. The run's conclusion is `cancelled`,
not `failure`, and no test output appears in the log, because the step runs
before the suite starts.

Observed on one commit:

```
py3.12   apt started 07:09:04, last output 07:09:42, killed 13:09:27
py3.13   apt started 07:09:06, last output 07:09:44, killed 13:09:27
verify   apt started 07:09:52, last output 07:10:31, killed 13:10:05
```

`py3.11` passed in four minutes on the same commit and the same image — it
cleared apt before the mirror went quiet — and the identical suite passed on
all three interpreters at the previous head. So the code was never at fault.

That single occurrence burned roughly 18 job-hours. It has recurred several
times across different pull requests, including runs from the dependency bot,
which is what makes it recognisable as infrastructure rather than a change.

Two costs beyond the wasted minutes:

1. **It reads as a real failure.** A reviewer seeing three red checks has to
   open the logs and reason about job conclusions to discover nothing ran.
2. **It is unbounded.** The step has no `timeout-minutes`, so a stalled mirror
   consumes a full job slot rather than failing fast.

Worth considering, roughly in order of value:

- A step-level or job-level `timeout-minutes` so a stall fails in minutes.
- A retry around the install.
- Caching the system packages, or using a runner image that already has them,
  so the common path does not hit a mirror at all.

The first alone would remove most of the harm.

md-mt:
Fixes #183.

## What was happening

Run [32226489311](https://github.com/md-mt/termproof/actions/runs/32226489311) attempt 1, the one in the issue:

| job | apt started | apt ended | duration |
|---|---|---|---|
| py3.12 | 07:09:04 | 13:09:24 | **6h00m** |
| py3.13 | 07:09:06 | 13:09:25 | **6h00m** |
| verify | 07:09:11 | 13:09:50 | **6h00m** |
| py3.11 | 07:09:11 | 07:11:33 | 2m22s ✅ |

18 job-hours from one commit. Attempt 2, re-run ten hours later on the same tree, cleared apt in 18–52s. The code was never at fault.

This is not confined to the kills. Across **214 apt-step samples** from CI history, the step *succeeded* at up to **43.4 minutes**:

| | median | p90 | p95 | p99 | max |
|---|---|---|---|---|---|
| apt step (n=197 successes) | **28s** | 498s | 766s | 2264s | **2601s** |

The six-hour kills are the tail of that distribution, not a separate phenomenon.

## What ffmpeg was actually for

I checked before bounding it, and the install was **redundant in all three places it appeared**.

`termproof.evidence.find_ffmpeg` prefers an ffmpeg on `PATH` and falls back to `imageio_ffmpeg.get_ffmpeg_exe()`. `imageio-ffmpeg` is a **hard runtime dependency** in `pyproject.toml`, and its `manylinux2014_x86_64` wheel is 29.5 MB because it carries the binary. `uv sync` had already put a working ffmpeg on every one of those runners.

Verified on a machine with no system ffmpeg at all:

```
system ffmpeg on PATH: None
find_ffmpeg()        -> .../imageio_ffmpeg/binaries/ffmpeg-macos-aarch64-v7.1
                        ffmpeg version 7.1
```

and it does the real job, not just answer `-version` — a full `render_mp4` through the actual code path:

```
Stream #0:0 Video: h264 (High) (avc1), yuv420p(progressive), 790x560, 12 fps
```

Per job, by instrumenting `subprocess` across the whole suite:

| job | invokes ffmpeg? | needed apt? |
|---|---|---|
| `test` (×3 interpreters) | once, and it resolved to the **bundled** binary | no |
| `verify` | yes, renders evidence video | no — bundled binary is sufficient |
| `release` | yes, renders evidence video | no — same |

741 tests passed in 73s with no system ffmpeg present, matching CI's own 67–87s.

So all three apt steps are **deleted** rather than retried. The two jobs that render video now assert the bundled binary resolves and runs — no network, and it fails there with a clear message instead of thirty steps later inside a render.

## The three layers

**1. Bound it.** All **25 jobs across 12 workflows** now declare `timeout-minutes` — not one did before, so every job in this repo could burn six hours. Every package-install step declares its own too.

Sized off **404 successful job runs**. The two apt jobs were the *only* ones with a pathological tail; every other job maxes at 13.3m:

| job | median | p95 | max | chosen |
|---|---|---|---|---|
| Unit tests and coverage | 2.5m | 27.3m | 44.9m | **20m** |
| Build, verify TUI evidence | 3.5m | 10.3m | 18.9m | **30m** |
| CI (Rust) lint/test | 10.4m | 12.8m | 13.3m | **45m** |
| Docker Image (Rust) | 3.0m | 3.6m | 3.8m | **30m** |

With the mirror gone the unit-test job's own steps total under two minutes, so 20m is roughly tenfold headroom. **These are backstops, not budgets** — a timeout tight enough to trip on ordinary variance would trade one flake for another.

**2. Retry it.** `cargo install` is the remaining step that reaches the network, so it runs under `.github/scripts/retry.sh` (3 attempts, 480s each, linear backoff).

The per-attempt bound is the whole point. **A bare `timeout-minutes:` cannot make a retry useful against a stall** — the first attempt hangs, consumes the entire step budget, and the step dies having never reached attempt two. The bound has to sit on each attempt. `timeout-minutes:` stays on the step as the outer backstop.

**3. Off the mirror.** Covered above.

## Evidence that it works

`test_retry_script.py`, run against **both** timeout implementations (coreutils `timeout`, and the shell watchdog the macOS runners fall back to):

- **retry recovers from a single simulated failure** — command fails once then succeeds; exit 0, log shows `attempt 1/3 exited 1` → `succeeded on attempt 2/3`
- **the timeout fires on a hang and still retries** — attempt 1 sleeps **600s**, per-attempt limit 3s; whole invocation finishes in **4s** with exit 0. This is the #183 failure mode, recovered.
- **a permanent stall exhausts its attempts** — exits **124** in ~5s rather than running to the ceiling
- **a permanent error propagates its status** — exits 7, not 124
- a self-check asserts the sanitised PATH really hides `timeout`, so the watchdog branch cannot silently test the same path twice

Locally green: **757 tests** (741 + 16 new), `ruff`, `mypy`, the stdlib-only job, version drift, schema drift. The new `Confirm ffmpeg` step body was extracted from the YAML and executed verbatim.

## Guards so it cannot come back

- `test_ci_timeouts.py` — fails if any job or install step is added without a bound, and carries the measured table the numbers came from
- `test_ffmpeg_needs_no_mirror.py` — keeps `imageio-ffmpeg` a hard dependency and fails if `apt-get` reappears in any workflow
- `test_retry_script.py` — pins the retry and the per-attempt bound on both implementations

## Notes for the reviewer

- Affects CI only. Neither published artifact changes.
- No repository settings, secrets or variables touched.
- Two follow-ups I deliberately left out of scope: the `test` job invokes `agg` **zero** times, so its `Install agg` step may be removable (it is a 0s cache hit today); and `README.md` still tells consumers to `apt-get install ffmpeg`, which the bundled dependency arguably makes stale — that is claims-sweep territory, and #185 is already in that area.

### DIFF
--- a/.github/scripts/retry.sh
+++ b/.github/scripts/retry.sh
@@ -1,0 +1,116 @@
+#!/usr/bin/env bash
+#
+# Run a command with a per-attempt timeout and bounded retries.
+#
+# A step-level `timeout-minutes:` bounds the whole step, which is enough to
+# stop a stall burning the six-hour job ceiling but not enough to recover from
+# one: the first attempt hangs, eats the entire budget, and the step dies with
+# no retry ever reached. The timeout has to sit on each *attempt* for a retry
+# to be worth having. That is what this exists for — `timeout-minutes:` stays
+# on the step as the outer backstop.
+#
+# Usage:
+#   retry.sh [--attempts N] [--timeout SECONDS] [--delay SECONDS] -- cmd [args...]
+#
+# Defaults: 3 attempts, 300s per attempt, 15s base delay (linear backoff).
+# Exit status is the last attempt's, or 124 if the last attempt timed out.
+
+set -uo pipefail
+
+attempts=3
+per_attempt_timeout=300
+delay=15
+
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --attempts) attempts="$2"; shift 2 ;;
+    --timeout) per_attempt_timeout="$2"; shift 2 ;;
+    --delay) delay="$2"; shift 2 ;;
+    --) shift; break ;;
+    *) echo "retry.sh: unknown option '$1'" >&2; exit 2 ;;
+  esac
+done
+
+if [ $# -eq 0 ]; then
+  echo "retry.sh: no command given" >&2
+  exit 2
+fi
+
+# `timeout` is coreutils and is absent on the macOS runners, where Homebrew
+# coreutils installs it as `gtimeout`. Fall back to a shell watchdog so this
+# behaves the same on every runner we use rather than silently running
+# unbounded on one of them.
+run_with_timeout() {
+  local limit="$1"; shift
+  # `-k 5` escalates to SIGKILL five seconds after SIGTERM, matching what the
+  # watchdog below does, so a command that ignores TERM cannot outlive its
+  # bound on either path.
+  if command -v timeout >/dev/null 2>&1; then
+    timeout -k 5 "$limit" "$@"
+    return $?
+  fi
+  if command -v gtimeout >/dev/null 2>&1; then
+    gtimeout -k 5 "$limit" "$@"
+    return $?
+  fi
+
+  # The watchdog marks the file immediately before signalling. Liveness is not
+  # usable as the signal here: after `kill -TERM` the watchdog is still in its
+  # grace sleep, so it looks alive while the command it just killed reports
+  # 143. The marker distinguishes "timed out" from "died on its own" exactly.
+  local marker
+  marker="$(mktemp)"
+  rm -f "$marker"
+
+  "$@" &
+  local cmd_pid=$!
+  (
+    sleep "$limit"
+    : > "$marker"
+    kill -TERM "$cmd_pid" 2>/dev/null && sleep 5 && kill -KILL "$cmd_pid" 2>/dev/null
+  ) >/dev/null 2>&1 &
+  local watchdog_pid=$!
+
+  local status=0
+  wait "$cmd_pid" 2>/dev/null || status=$?
+  kill -TERM "$watchdog_pid" 2>/dev/null
+  wait "$watchdog_pid" 2>/dev/null || true
+  # Report 124 the way coreutils `timeout` does, so both paths agree.
+  if [ -f "$marker" ]; then
+    status=124
+  fi
+  rm -f "$marker"
+  return "$status"
+}
+
+status=0
+attempt=1
+while [ "$attempt" -le "$attempts" ]; do
+  status=0
+  run_with_timeout "$per_attempt_timeout" "$@" || status=$?
+
+  if [ "$status" -eq 0 ]; then
+    if [ "$attempt" -gt 1 ]; then
+      echo "retry.sh: succeeded on attempt ${attempt}/${attempts}"
+    fi
+    exit 0
+  fi
+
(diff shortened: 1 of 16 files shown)


md-mt:
## Measured on this PR's own CI run

Run [32311296372](https://github.com/md-mt/termproof/actions/runs/32311296372) — the numbers in the description were local; these are from the runners.

**The apt step no longer exists, and the variance went with it.**

| job | before (median / p95 / max, n=404) | this run |
|---|---|---|
| Unit tests py3.11 | 2.5m / 27.3m / 44.9m | **125s** |
| Unit tests py3.12 | ″ | **119s** |
| Unit tests py3.13 | ″ | **116s** |
| Build, verify TUI evidence | 3.5m / 10.3m / 18.9m | **191s** |

The three test jobs now land within 9 seconds of each other. Previously the same three jobs on one commit ranged 24s → 2601s on the install step alone.

**No job lost a dependency it uses.** The `verify` job resolved the bundled binary on the runner:

```
/home/runner/work/termproof/termproof/python/.venv/lib/python3.12/site-packages/imageio_ffmpeg/binaries/ffmpeg-linux-x86_64-v7.0.2
ffmpeg version 7.0.2-static
```

and then `Run generic and Pi agent UI verification` — the step that actually renders evidence video through `agg` + ffmpeg — passed in 63s. `Confirm ffmpeg resolves without a package mirror` cost 1s.

Worth noting `imageio-ffmpeg` downloaded in **0.7s** as part of the dependency sync that happened anyway, since it is a hard dependency. The apt install was never buying anything.

All 24 checks: 21 pass, 3 skip, 0 fail.

One honest caveat: `Run unit tests with coverage` rose from ~80s to ~107s, because `test_retry_script.py` deliberately sleeps while exercising the per-attempt timeout on both implementations. That is the new tests' own cost, not a regression, and it sits far inside the 20m bound.

PR #192, opened 2026-08-20
docs: say plainly that capture_text's signature differs across implementations

md-mt:
## What

`EvidenceCollector.capture_text` takes the `CaptureKind` positionally in Rust and defaults it to `CaptureKind.CHECKPOINT` in Python. Neither is wrong — each is idiomatic where it sits — but neither README mentioned the method at all, and everywhere else the two surfaces mirror each other, so a reader would have carried that symmetry here and been wrong.

Both READMEs now name the difference, in the same words, and give both spellings:

- `rust/crates/termproof/README.md` gains an `EvidenceCollector` bullet in the `evidence` surface list.
- `python/README.md` gains a short **Evidence collector** section — it had no library-API section at all, and the collector is new in 0.4.0.

## Why documentation rather than an API change

The other option was a Rust `capture_text_checkpoint` helper. That gives Rust the convenience Python gets from a default argument, but it does so under a name Python has no reason to grow — trading a signature asymmetry for a *name* asymmetry, plus a new abstraction, for a difference the feedback itself calls "not wrong". Documenting it is the smaller honest fix.

## Parity

No public surface was added, so the conformance pair (`conformance/probe_evidence_manifest.py` / `rust/crates/termproof/tests/differential_evidence_manifest.rs`) needs no extension — both already exercise `capture_text` with an explicit kind, which is the spelling that is identical on both sides.

The Python default it documents was exercised but never asserted; `test_text_can_be_captured_without_a_source` now asserts it, so the documented behaviour has a test holding it.

## Verification

`uv run --frozen python -m unittest tests.test_collector` — 21 tests, OK. No Rust code changed. No version bumped; the entry sits under the existing `[Unreleased]` → `Docs` convention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,18 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
 
 ## [Unreleased]
 
-Nothing yet.
+### Docs
+
+- **Both READMEs now say that `EvidenceCollector.capture_text` is the one
+  collector signature the two implementations do not share.** Rust takes the
+  `CaptureKind` positionally and Python defaults it to
+  `CaptureKind.CHECKPOINT`; each is idiomatic where it sits, and neither README
+  mentioned the method at all, so a reader carrying the symmetry the rest of
+  the surface has would have carried it here too and been wrong. Reported by a
+  consumer running both implementations against the same collector code. The
+  signatures are unchanged — the Python default is now covered by an assertion
+  in `test_text_can_be_captured_without_a_source` rather than only being
+  exercised.
 
 ## [0.4.0] — 2026-08-19
 
--- a/python/README.md
+++ b/python/README.md
@@ -96,6 +96,23 @@ Each run writes under `.termproof/runs/<run-id>/` (or the `--out` you provide):
 - `report.md` — per-run review summary
 - `latest-report.md` — aggregate report for multi-recipe runs
 
+## Evidence collector
+
+`termproof.collector.EvidenceCollector` is the ordered step model for a caller
+driving its own run rather than a recipe. `capture` and `capture_failure` pull
+from a `ScreenSource`; `capture_text` records a screen you already hold — text
+recovered from a log, or a golden file in a test:
+
+```python
+collector.capture_text("from-log", recovered)
+collector.capture_text("post-mortem", last_screen, CaptureKind.FAILURE)
+```
+
+**This is the one collector signature the two implementations do not share.**
+Here `kind` defaults to `CaptureKind.CHECKPOINT`; Rust takes it positionally,
+as `capture_text(label, screen, CaptureKind::Checkpoint)`. The meaning, the
+ordering and the resulting manifest are identical.
+
 ## Recipe example
 
 ```json
--- a/python/tests/test_collector.py
+++ b/python/tests/test_collector.py
@@ -127,6 +127,9 @@ class CaptureTest(unittest.TestCase):
         self.assertEqual(3, len(steps))
         self.assertEqual(1, steps[1].index)
         self.assertEqual("recovered", steps[1].screen)
+        # The default the READMEs document as the difference from Rust, which
+        # takes the kind positionally.
+        self.assertEqual(CaptureKind.CHECKPOINT, steps[1].kind)
         self.assertEqual(CaptureKind.FAILURE, steps[2].kind)
         self.assertIsNone(steps[2].raw_output)
 
--- a/rust/crates/termproof/README.md
+++ b/rust/crates/termproof/README.md
@@ -137,6 +137,14 @@ a comment on each one that sits above the oldest workable version.
 
 ### `evidence` — screenshots, video, reports
 
+- `EvidenceCollector` — an ordered step model for a caller driving its own run.
+  `capture` and `capture_failure` pull from a `ScreenSource`;
+  `capture_text(label, screen, kind)` records a screen the caller already
+  holds, and takes the `CaptureKind` positionally. **This is the one collector
+  signature the two implementations do not share**: Python's is
+  `capture_text(label, screen, kind=CaptureKind.CHECKPOINT)`, so the Rust call
+  spells out `CaptureKind::Checkpoint` where the Python one may omit it. The
+  meaning, the ordering and the resulting manifest are identical.
 - `render_svg` / `render_by_extension` — plain screen text to an SVG, in
   default colours.
 - `ScreenshotRenderer` — an attributed grid to a PNG, with the colours the


md-mt:
# PR 192 — docs: say plainly that capture_text's signature differs across implementations

**VERDICT:** MERGE WITH NITS
**MERGE RISK:** low — no behaviour changes; the only exposure is one over-absolute sentence in the new prose that could send a reader past a second, smaller asymmetry it implicitly denies.
**CI:** 23 checks, 20 pass / 3 skip / 0 fail, and it is meaningful coverage. Verified fresh, not stale: the CI evidence comment names head `fb465f8`, which is GitHub's ephemeral merge commit of PR head `41068b4` into base `f033dcc` (`git rev-parse pr192merge` → parents `f033dcc`, `41068b4`). So CI ran on exactly the head reviewed here, already merged into current main. The relevant gates are `Unit tests and coverage (py3.11/3.12/3.13)`, `Lint, type-check and test (ubuntu + macos)`, and `Public API compatibility (cargo semver-checks)`. The 3 skips are Pages-deploy and publish jobs, correctly gated off.
**STALENESS:** None. Head `41068b4` is a direct child of current `origin/main` (`f033dcc`). `git merge-base --is-ancestor origin/main pr192` passes — no rebase needed, applies cleanly.

## Summary

The PR documents a real asymmetry and documents it accurately. I read both `capture_text`
implementations at this PR's head rather than trusting the body, and the claim holds exactly as
written: Rust requires the `CaptureKind` positionally, Python defaults it to
`CaptureKind.CHECKPOINT`. The one test assertion it adds does have teeth — I proved it by mutating
the default and watching it go red. The parity claim about the conformance pair is also true. The
single thing I would change before merge is a quantifier: the load-bearing sentence claims
`capture_text` is *the one* collector signature that differs, and construction is a second one.

Diff is 4 files, 40 insertions, 1 deletion. Nothing outside docs, changelog and one assertion.

## Findings

### 1. minor — "the one collector signature" is a false absolute; construction also diverges

- **location:** `python/README.md:111`, `rust/crates/termproof/README.md:143-144`, echoed in `CHANGELOG.md:46-47`
- **verified**, by enumerating the full public surface of both `EvidenceCollector` types at this head.

Both READMEs state, in bold: **"This is the one collector signature the two implementations do not
share."** Construction is a second one, and it has the same root cause the PR itself names — Rust
has no default arguments:

| | Rust | Python |
|---|---|---|
| default grid | `EvidenceCollector::new()` (`collector.rs:320`) | `EvidenceCollector()` |
| custom grid | `EvidenceCollector::with_grid(columns, rows)` (`collector.rs:328`) | `EvidenceCollector(columns, rows)` (`collector.py:360`) |

Rust needs a separately *named* constructor for the case Python expresses with optional arguments.
That is precisely the shape of divergence this PR exists to warn about.

Concrete failure case: a reader who needs a non-default grid reads the bolded sentence, concludes
`capture_text` is the only place the two surfaces part company, and writes
`EvidenceCollector::new(100, 30)` in Rust (wrong arity — `new` takes none) or
`EvidenceCollector.with_grid(100, 30)` in Python (`AttributeError`). The doc did not merely fail to
warn them; it actively told them there was nothing else to watch for. For a PR whose entire value is
precision about asymmetry, an over-strong quantifier is the one defect that matters.

What I would do: drop the absolute. "This is the one *capture* signature the two implementations do
not share" is true and costs one word. Alternatively keep "the one collector signature" and add a
half-sentence noting `with_grid` as the same idiom in the constructor — that is arguably the better
doc, since it names the *pattern* (Rust spells out what Python defaults) rather than one instance of
it. Either is a prose-only change; no code moves.

I checked the rest of the surface and found nothing further: `capture`, `capture_failure`,
`attach_recording`, `steps`, `recordings` and `publish` all mirror; Rust's `len()`/`is_empty()`
against Python's `__len__` is idiomatic mirroring, not divergence; and `DEFAULT_COLUMNS` is 120 on
both sides (`attributed.py:32`, `attributed.rs:70`), so the grid *values* agree even though the
construction spelling does not.

### 2. nit — the new Python snippet is the file's only `python` block and omits its imports

- **location:** `python/README.md:106-109`
- **verified** (block inventory of the file; snippet executed).

The block opens mid-flow on `collector.capture_text(...)` and uses `CaptureKind` unqualified, with
no `from termproof.collector import EvidenceCollector, CaptureKind` above it. It is the only
` ```python ` block in the file (every other block is bash, json or yaml), and `python/README.md` is
the PyPI `long_description` (`pyproject.toml:5`), so it is the one Python example a package-page
reader sees. Two import lines would make it copy-pasteable. Marking this as taste, not defect — the
fragment reads clearly in context.

## What is good

Each of these I checked against the tree rather than the PR body.

- **The documented claim is true.** Rust: `pub fn capture_text(&mut self, label: &str, screen: &str, kind: CaptureKind)` — `rust/crates/termproof/src/evidence/collector.rs:359`, positional and required. Python: `def capture_text(self, label: str, screen: str, kind: CaptureKind = CaptureKind.CHECKPOINT)` — `python/termproof/collector.py:376-378`. Both READMEs describe exactly this. The failure mode the brief warned about — a confidently wrong doc — did not happen.

- **The new assertion can fail, and I proved it rather than read it.** I changed the Python default from `CHECKPOINT` to `FAILURE` in the source and reran: `test_text_can_be_captured_without_a_source` failed at `python/tests/test_collector.py:137` with `<CaptureKind.CHECKPOINT: 'checkpoint'> != <CaptureKind.FAILURE: 'failure'>`. Reverted, 21 tests OK again. Worth one caveat for the maintainer: the same mutation applied to pre-PR `main` (`f033dcc`) *already* failed one test — `test_the_manifest_is_written_and_reads_back` at line 309, which reaches the default indirectly through the written manifest. So the default was not previously unguarded; what the PR adds is a *direct, named* assertion at the point of the behaviour, with a comment tying it to the READMEs. The CHANGELOG's wording ("covered by an assertion ... rather than only being exercised") is literally accurate, and the direct assertion is the better guard because it localises the failure. But it is a clarity win, not new regression coverage.

- **The parity claim is true.** Both halves of the conformance pair call `capture_text` with an explicit kind at four mirrored call sites each — `conformance/probe_evidence_manifest.py:94,97,105,106` against `rust/crates/termproof/tests/differential_evidence_manifest.rs:134,137,145,146`, same labels, same order, same kinds. No public surface was added, so no conformance extension is owed. Correct as stated.

- **The new README section describes nothing that does not exist.** I executed the snippet verbatim: `capture_text("from-log", recovered)` → step 0 `CaptureKind.CHECKPOINT`, `capture_text("post-mortem", last_screen, CaptureKind.FAILURE)` → step 1 `CaptureKind.FAILURE`, `raw_output=None` on both — exactly what the prose promises. The module path `termproof.collector.EvidenceCollector` is the right one: `python/termproof/__init__.py` does not re-export the collector symbols, so the qualified path is the only working spelling and the README uses it. `ScreenSource` and `CaptureKind` are both in `collector.__all__`.

- **The Rust bullet is filed correctly.** `EvidenceCollector` is re-exported from `termproof::evidence` (`rust/crates/termproof/src/evidence/mod.rs:34`), so the `evidence` surface list is the right home, and the bullet's length and style match its neighbours in that list.

- **Choosing docs over a `capture_text_checkpoint` helper was the right call**, and for a reason stronger than the PR gives itself: a Rust helper would need `cargo semver-checks` to carry it forever as public API, in exchange for hiding a difference that the language guarantees is discoverable at compile time. Rust *cannot* grow a default argument, so this asymmetry can only ever be closed from the Python side — and if someone removed the Python default, `test_text_can_be_captured_without_a_source` breaks at the call site, not just the assertion. Both halves of the documented claim are anchored: Rust by the compiler and semver-checks, Python by this test. That is a well-defended doc.

- **`### Docs` is an existing changelog convention** (`CHANGELOG.md:204`), and unqualified rather than `Python —`/`Rust —` is right here because the change spans both trees.

## Merge-order interaction with PR 193

Both PRs branch from the same base (`f033dcc`). I trial-merged them rather than reasoning about it.

- **`python/tests/test_collector.py` does NOT conflict**, textually or semantically. 193's hunks land at lines 11, 24 and 398; 192's assertion lands at 130-137. `git merge` reports `Auto-merging python/tests/test_collector.py` with no conflict. Running the full Python suite on the merged tree: **793 tests, OK.** 192's assertion survives the merge intact and still passes against 193's refactored dedup.
- **`CHANGELOG.md` DOES conflict**, guaranteed and trivially. Both replace the same single `-Nothing yet.` line under `## [Unreleased]` — 192 with a `### Docs` block, 193 with a `### Python — Changed` block. Whichever lands second resolves by keeping both sections.
- **Recommendation: land 192 first.** It is 40 lines of prose plus one assertion with nothing to re-verify, so the second lander's conflict resolution is cheaper if it is 193 — that PR already needs its 793-test suite rerun post-rebase for its dedup refactor, and folding a three-line changelog fixup into a run it must do anyway costs nothing. The reverse order forces a rerun on 192 that its content does not otherwise require. This is a convenience ordering, not a correctness constraint; either order is safe.

## Outside this PR

- The CI evidence comment shows seven recipes as `SKIP -> PASS` behavioural deltas. That is an artifact of the base commit producing no TermProof report at all ("Before: ... No TermProof report was generated"), not anything this docs PR did. Flagging it only so the delta list is not misread as a docs PR moving seven behaviours.
- `python/tests/test_public_claims.py` is a banned-phrase sweep for *withdrawn* claims — it cannot catch a newly-written claim that is wrong, which is the class this PR is in. Nothing in CI validates README code snippets against the real API (no `#![doc = include_str!("../README.md")]` in `rust/crates/termproof/src/lib.rs`, no doctest runner over `python/README.md`). Not this PR's job to build that, and here the exposure is small because the API claims are anchored by the compiler and the new assertion. Worth knowing the guard does not exist.

## What I could not verify

Nothing material. I did not run the Rust test suite, because the diff touches no Rust code — only `rust/crates/termproof/README.md`, which is not included as crate documentation, so it is not compiled as doctests. CI's Rust jobs passed on the merge commit regardless.

---

### Commands run (worktree scratch, nothing kept or pushed)

```
git fetch origin pull/192/head:pr192 && git checkout pr192          # 41068b4
uv run --frozen python -m unittest tests.test_collector             # 21 tests, OK
# mutate default CHECKPOINT -> FAILURE in python/termproof/collector.py
uv run --frozen python -m unittest tests.test_collector             # FAILED (failures=2), incl. line 137
# same mutation applied to pre-PR f033dcc python tree
uv run --frozen python -m unittest tests.test_collector             # FAILED (failures=1) -> prior indirect coverage
git checkout pr192 -- python/                                       # restored, 21 tests OK
git merge --no-commit --no-ff pr193                                 # only CHANGELOG.md conflicts
uv run --frozen python -m unittest discover -s tests                # 793 tests, OK
git merge --abort; git checkout pr192; git branch -D mergeprobe     # cleaned
git rev-parse pr192merge                                            # fb465f8 = merge(41068b4, f033dcc)
```

Worktree left at `pr192` (`41068b4`), clean. No commits, no pushes, no PR comments, no labels.

PR #193, opened 2026-08-20
evidence: one step-screenshot dedup for the Python package, not two

md-mt:
## What

`python/termproof/collector.py` and `python/termproof/evidence.py` each carried
their own copy of the step-screenshot dedup. Both now ask
`termproof.dedup.Deduper`, the single copy of the rule in the package.

The collector is not decorative. It is the imperative capture path, and it now
shares the one piece of logic it genuinely had in common with the declarative
one. The queued `record_session` work can build on `EvidenceCollector` as the
real path.

## Why not `evidence` → `EvidenceCollector.publish` verbatim

The brief prefers wiring `evidence`'s module-level functions on top of
`EvidenceCollector` "so the step-screenshot dedup exists exactly once". That
purpose is met. Routing `_render_step_screens` through `publish()` itself is
not, because the two disagree on six things that are public behaviour and that
existing tests pin:

| | `render_artifacts` | `publish` |
|---|---|---|
| Filenames | `steps/NN-name.svg`, 1-based | `step-NN-label.png`, 0-based, label truncated to 40 |
| Step text | `step.screen + "\n"` | verbatim, no trailing newline |
| Manifest | `steps-manifest.json`, `{step, screenshot, unchanged_from_previous}`, screenshot is a bare name | `evidence.json`, `same_as: {index, label}`, screenshot is a full path |
| Dedup | off unless `evidence.dedup_step_screenshots` | always on |
| Renderer | text-only `render(text, path, cols, rows)` still supported for third-party renderers | requires `render_attributed` |
| Render failure | propagates, fails the run | recorded on the step as `error` |

Plus `publish` needs a `RunIdentity`, which `_render_step_screens` has no
recipe/renderer/run-id to build.

Making `publish` absorb those would mean six strategy knobs on the collector,
which is a worse collector for the sake of a wrapper — and every knob is a
chance to change the `evidence.json` a downstream consumer already reads. The
genuinely shared part is the dedup, and that is what got extracted.

This is not the documentation-only fallback: the duplicated logic is gone, not
annotated.

## Rust parity

**Rust does not have this split, and needs no change.**

- `rust/crates/termproof/src/evidence/dedup.rs` already holds a `Deduper`, and
  `evidence/collector.rs:428` is its only caller. Rust factored the rule out at
  consolidation.
- `evidence::report` is a Markdown/JUnit generator over `RunResult`. It has no
  step-screenshot rendering, no `steps/` layout, and no `unchanged_from_previous`
  — `grep` for those turns up Python only. So there was no second Rust copy to
  reconcile.

`python/termproof/dedup.py` mirrors `dedup.rs`, including its API
(`check` / `forget` / `last_rendered`). It sits at top level rather than inside
`termproof/evidence.py` because the Python package flattens the Rust
`evidence::` namespace into top-level modules — the same mapping that puts
`crate::evidence::collector` at `termproof/collector.py` and
`crate::terminal::attributed` at `termproof/attributed.py`.

The wider Python/Rust divergence — Rust has no `render_artifacts` equivalent at
all — is pre-existing and out of scope here.

## Conformance

No new cross-implementation surface, so the conformance pair is unextended, and
that is deliberate: it compares *documents*, and this change produces the same
documents. `probe_evidence_manifest.py` already exercises the dedup (it captures
`menu-open` twice, and `styled`/`unstyled` for the colour-only case), so the
shared rule is already differentially covered through `evidence.json`.

`Deduper`'s own behaviour is pinned by mirrored unit tests instead:
`python/tests/test_dedup.py` transcribes `dedup.rs`'s test module case for case.

## Behaviour

Unchanged, and not by weakening anything:

- 793/793 Python tests pass, including `test_evidence_config.py`'s
  `StepScreenshotDedupTest` and `test_step_attributed_screens.py`'s dedup cases,
  untouched.
- `conformance/probe_evidence_manifest.py` reproduces
  `conformance/corpus/evidence_manifest.expected.json` byte for byte (`diff`
  clean), so the published `evidence.json` and every file it points at are
  identical.
- `ruff check .` and `mypy termproof` clean.

The one place the code got tighter rather than equivalent: the old `publish`
had a dead `previous_fingerprint = None` in its no-renderer branch, overwritten
two lines later. Its effect was already nil (the reuse branch was separately
guarded on `last_rendered is not None`), and the no-renderer path is covered by
`test_text_is_written_even_with_no_renderer`.

## Tests

- `python/tests/test_dedup.py` (new) — mirrors `dedup.rs`'s tests: first screen
  renders, run of identical screens all point at the *first* not the neighbour,
  colour-only change is a change, lookback is one step only, `forget` on a
  failed render.
- `python/tests/test_collector.py` — new `SharedDedupTest` covers the delegation
  path on both sides by substituting the deduper, so it fails if either module
  re-grows its own copy *even if that copy is correct*:
  - collector with a never-matching deduper renders two identical screens twice;
  - `render_artifacts` with an always-matching deduper reports a sentinel image
    for two *different* screens — a verdict nothing could reach on its own;
  - a failed render calls `Deduper.forget`;
  - dedup off never consults the deduper at all.

## Not done

No version bump. `CHANGELOG.md` entry is under `[Unreleased]` → `Python — Changed`.

Local note: Rust test binaries are SIGKILLed on startup in this worktree
(sandbox/Gatekeeper), on a clean tree as well as this one, so
`differential_evidence_manifest` was verified from the oracle side only — the
Python half, which is the half this PR could have moved, reproduces the corpus
exactly. CI runs the Rust half.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,30 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   in `test_text_can_be_captured_without_a_source` rather than only being
   exercised.
 
+### Python — Changed
+
+- **evidence:** the step-screenshot dedup was written out twice — once in
+  `EvidenceCollector.publish` and once in `evidence._render_step_screens` —
+  and the two copies shared no code, so nothing stopped them drifting on what
+  counts as a changed screen. Both now ask the new `termproof.dedup.Deduper`,
+  which is the single copy of the rule. `termproof.dedup` mirrors
+  `termproof::evidence::dedup`, where the Rust implementation has kept the rule
+  in one place since the consolidation; the Python package flattens the Rust
+  `evidence::` namespace into top-level modules, so it lands beside
+  `termproof.collector` rather than inside `termproof.evidence`.
+
+  Nothing observable moved. `render_artifacts` writes the same `steps/` layout
+  and the same `steps-manifest.json`, still gated on
+  `evidence.dedup_step_screenshots` and still off by default; `publish` writes
+  the same `evidence.json`, byte-for-byte — `conformance/probe_evidence_manifest.py`
+  reproduces the committed corpus unchanged. The two file layouts are still
+  separate, and folding one into the other is still a compatibility question of
+  its own rather than something to smuggle into a dedup fix.
+
+  Both module docstrings now say plainly which type owns what:
+  `EvidenceCollector` owns capture-while-running, `render_artifacts` owns
+  rendering a finished `RunResult`, and neither owns the dedup.
+
 ## [0.4.0] — 2026-08-19
 
 ### Rust — Changed
--- a/python/termproof/collector.py
+++ b/python/termproof/collector.py
@@ -24,9 +24,18 @@ picture cannot answer that question.
 Which of this and :mod:`termproof.evidence` to use
 --------------------------------------------------
 
-Both render screens and both dedupe them, and they are **not** layered on one
-another. They answer different questions and produce different documents, so
-the split is deliberate rather than pending cleanup:
+Who owns what:
+
+* :class:`EvidenceCollector` owns capture — deciding *while a run is going on*
+  which moments are worth keeping — and publishing what it captured.
+* :func:`termproof.evidence.render_artifacts` owns rendering the artifacts a
+  finished :class:`RunResult` already carries, whose step list the recipe fixed
+  before the run started.
+* :class:`termproof.dedup.Deduper` owns the "has this screen changed?" rule, for
+  both of them. Neither implements it; there is one copy in the package.
+
+They still write different documents, and that is why one is not a thin wrapper
+over the other:
 
 ===============  =================================  ===============================
                  :mod:`termproof.evidence`          this module
@@ -35,8 +44,7 @@ Driven by        a finished :class:`RunResult`      a caller, while it runs
 Step list        the recipe's, fixed in advance     whatever the caller captured
 Writes           ``final.*``, ``steps/``,           ``step-NN-label.*`` and
                  ``steps-manifest.json``, video     ``evidence.json``
-Dedup fingerprint the step's grid, or one parsed    the grid the source reported,
-                 from its ANSI text                 or one built from plain lines
+Dedup verdict    :class:`~termproof.dedup.Deduper`  :class:`~termproof.dedup.Deduper`
 Dedup records    ``unchanged_from_previous``        ``same_as: {index, label}``
 Dedup is         off unless configured              always on
 A render failure fails the run                      is recorded on the step
@@ -48,10 +56,10 @@ Use :mod:`termproof.evidence` for a declarative recipe run by
 this module when the caller decides what to capture as it goes, which a
 declarative recipe cannot express.
 
-Consolidating the two would mean changing the file layout
+Folding one of the two *file layouts* into the other is a different change from
+sharing the dedup, and not one to smuggle in here: it would move the paths
 :func:`termproof.evidence.render_artifacts` has always written, which every
-existing reader of a run directory depends on. That is a separate change with
-its own compatibility question, not a tidy-up to fold in here.
+existing reader of a run directory depends on.
 """
 
 from __future__ import annotations
@@ -63,6 +71,7 @@ from pathlib import Path
 from typing import Protocol
 
 from .attributed import DEFAULT_COLUMNS, DEFAULT_ROWS, AttributedScreen, attributed_screen_from_lines
+from .dedup import Deduper
 from .models import RunResult
 
 #: Schema version of the published manifest. Shared with the Rust
@@ -441,10 +450,12 @@ class EvidenceCollector:
         publisher.directory.mkdir(parents=True, exist_ok=True)
         extension = getattr(publisher.renderer, "extension", "png")
         published: list[PublishedStep] = []
-        # The image the dedup verdict refers to. Tracked as a step rather than
-        # a label because labels are caller-supplied and may repeat.
+        deduper = Deduper()
+        # The image the deduper's answer refers to. `Deduper.check` reports a
+        # label, and labels are caller-supplied and may repeat; tracking the
+        # step alongside is exact, because the deduper only ever looks back one
+        # rendered step and this is it.
         last_rendered: tuple[ReusedFrom, str, str | None] | None = None
-        previous_fingerprint: str | None = None
 
         for step in self._steps:
             stem = step.file_stem()
@@ -470,17 +481,17 @@ class EvidenceCollector:
                 raw_output=raw_output_path,
             )
 
-            # Fingerprint the grid the image is rendered from, not the text, so
-            # that a colour-only change counts as a change.
-            fingerprint = step.attributed.render_fingerprint()
-            if fingerprint == previous_fingerprint and last_rendered is not None:
-                source, image, url = last_rendered
-                entry.same_as = source
-                entry.screenshot = image
-                entry.url = url
-            elif publisher.renderer is None:
-                previous_fingerprint = None
-            else:
+            # The deduper fingerprints the grid the image is rendered from, not
+            # the text, so that a colour-only change counts as a change.
+            if deduper.check(step.label, step.attributed) is not None:
+                # Nothing to point at when there is no renderer: no image was
+                # ever made, so the step keeps its text and no screenshot.
+                if last_rendered is not None:
+                    source, image, url = last_rendered
+                    entry.same_as = source
+                    entry.screenshot = image
+                    entry.url = url
+            elif publisher.renderer is not None:
                 image_path = publisher.directory / f"{stem}.{extension}"
                 try:
                     publisher.renderer.render_attributed(
@@ -491,21 +502,19 @@ class EvidenceCollector:
                     )
                 except Exception as exc:  # noqa: BLE001 - best effort by contract
                     entry.error = str(exc)
-                    # Without this, the next identical screen is told to reuse
-                    # an image that was never produced.
+                    # `Deduper.forget`'s contract. Without it the next identical
+                    # screen is told to reuse an image that was never produced.
+                    deduper.forget()
                     last_rendered = None
-                    previous_fingerprint = None
-                    published.append(entry)
-                    continue
-                url = _upload(publisher, image_path)
-                entry.screenshot = str(image_path)
-                entry.url = url
-                last_rendered = (
-                    ReusedFrom(index=step.index, label=step.label),
-                    str(image_path),
-                    url,
-                )
-            previous_fingerprint = fingerprint
+                else:
+                    url = _upload(publisher, image_path)
+                    entry.screenshot = str(image_path)
+                    entry.url = url
+                    last_rendered = (
+                        ReusedFrom(index=step.index, label=step.label),
+                        str(image_path),
+                        url,
+                    )
             published.append(entry)
 
         manifest = EvidenceManifest(
--- a/python/termproof/dedup.py
+++ b/python/termproof/dedup.py
@@ -1,0 +1,115 @@
+"""Not re-rendering a screen that has not changed.
+
+A recipe captures evidence either side of an action, and plenty of actions turn
+out not to change the screen — a key that was already pressed, a wait that had
+nothing to wait for, a step that failed to do anything. Rendering and uploading
+a byte-identical image for each of those is pure cost, and a reviewer scrolling
+twelve identical images is worse off than one reading four distinct ones.
+
+:class:`Deduper` is where that rule lives, once, for the whole package. Both
+callers ask it the question and neither answers it itself:
+
+* :meth:`termproof.collector.EvidenceCollector.publish`, which publishes what a
+  caller captured while the run was going on;
+* :func:`termproof.evidence.render_artifacts`, which renders the steps a
+  finished :class:`~termproof.models.RunResult` already carries.
+
+The two write different documents — ``evidence.json`` with ``same_as`` against
+``steps-manifest.json`` with ``unchanged_from_previous`` — but "has this screen
+changed since the last one I rendered?" is one question, and answering it in two
+places is how the two answers drift apart.
+
+Mirrors ``termproof::evidence::dedup::Deduper``. The Python package flattens the
+Rust ``evidence::`` namespace into top-level modules, which is why
+``evidence::dedup`` lands here rather than inside :mod:`termproof.evidence`.
+
+Attributes count
+----------------
+
+Deduplication keys on
+:meth:`~termproof.attributed.AttributedScreen.render_fingerprint`, not on text.
+Two screens with the same characters but a different highlight are *different
+screenshots*, and a text-keyed cache would silently collapse them — losing
+exactly the frame where a selection moved.
+
+It only ever looks backwards one step
+-------------------------------------
+
+A run of identical screens all point at the first of them, but a screen that
+matches one from earlier — after something else in between — renders again. That
+is deliberate: evidence is read in order, and a caption saying "same as step 2"
+nine steps later costs the reader more than the image saves.
+
+.. code-block:: python
+
+    from termproof.attributed import attributed_screen_from_text
+    from termproof.dedup import Deduper
+
+    deduper = Deduper()
+    menu = attributed_screen_from_text("menu open", columns=20, rows=2)
+    same = attributed_screen_from_text("menu open", columns=20, rows=2)
+    gone = attributed_screen_from_text("menu closed", columns=20, rows=2)
+
+    assert deduper.check("opened", menu) is None
+    assert deduper.check("still-open", same) == "opened"
+    assert deduper.check("closed", gone) is None
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from .attributed import AttributedScreen
+
+
+@dataclass(frozen=True)
+class _Rendered:
+    """The screen a reuse verdict would point at."""
+
+    label: str
+    fingerprint: str
+
+
+class Deduper:
+    """Decides whether a screen needs rendering, or matches the previous one."""
+
+    def __init__(self) -> None:
+        self._previous: _Rendered | None = None
+
+    def check(self, label: str, screen: AttributedScreen) -> str | None:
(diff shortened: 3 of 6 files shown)


md-mt:
# PR 193 — evidence: one step-screenshot dedup for the Python package, not two

**VERDICT:** MERGE WITH NITS
**MERGE RISK:** low — I could not construct an input where the new code and `main` diverge; 22 hand-built edge cases and the cross-implementation corpus all reproduce byte for byte. The residual risk is not in this diff but in what CI *doesn't* check (below).
**CI:** 10/10 green, but narrower than the PR body claims. Only `CI (Python)` (9 jobs: unit tests on 3.11/3.12/3.13, lint+mypy, stdlib-only gate, three wheel builds, a real TUI evidence run) and `Docker Image (Python)` ran. **`CI (Rust)` did not run at all** — it is path-filtered to `rust/**`/`spec/**`/`conformance/**` and this PR touches none of them. The PR body's closing line "CI runs the Rust half" is false for this PR. I ran the Rust half myself instead, and it passes. Separately, *nothing* in CI reproduces the Python side of the conformance corpus — see "Outside this PR".
**STALENESS:** None. PR head `740982d` is parented directly on current `main` tip `f033dcc`; `git merge-tree` merges clean. CI ran on `da983f9`, the merge of the head into current main, so CI tested exactly the state that would land.

## Summary

`collector.py` and `evidence.py` each carried their own copy of the step-screenshot
dedup rule; both now delegate to a new `termproof.dedup.Deduper` that mirrors the
existing Rust `evidence::dedup::Deduper` line for line. The extraction is correct — I
could not find an input that behaves differently from `main` — and the refactored
`publish` is now a close structural mirror of `rust/.../collector.rs:427-500`, which is
a real convergence rather than a cosmetic one. The author declined the fuller
consolidation the task asked for and defended it with a six-row divergence table; **I
checked all six rows against the code and all six are accurate**, so the declination
rests on a true premise, not a false one. What is left is a wrong claim about Rust CI
coverage and a CHANGELOG that under-declares a new importable name.

## What I actually ran (verified vs. inferred)

Everything below marked **verified** I executed in this worktree at `740982d`.

| Claim | How checked | Result |
|---|---|---|
| 793/793 Python tests pass | **verified** — `uv run python -m unittest discover -s tests` | 793 tests, OK, 108s |
| `ruff check .` / `mypy termproof` clean | **verified** | "All checks passed!" / "no issues found in 42 source files" |
| Conformance corpus reproduces byte for byte | **verified** — ran `probe_evidence_manifest.py`, `diff` against `conformance/corpus/evidence_manifest.expected.json` | clean, zero diff |
| Rust half (`differential_evidence_manifest`) | **verified** — `cargo test --test differential_evidence_manifest` | 1 passed. Author's SIGKILL problem did **not** reproduce here |
| Rust `dedup` unit tests + doctest | **verified** — `cargo test --lib evidence::dedup`, `cargo test --doc dedup` | 9 passed, 1 doctest passed |
| Six-row divergence table | **verified** — all six rows against source, two of them empirically | all accurate (detail below) |
| `SharedDedupTest` bites | **verified** — re-grew a *correct* local dedup three different ways | goes red every time (detail below) |
| Behaviour unchanged | **verified** — built a 22-case differential battery, ran it on `main` and on the PR head | byte-identical output |
| No test weakened | **verified** — `git diff --numstat` on `python/tests/`: 159+/1− and 99+/0− | the single removed line is `-from termproof.models import RunResult`, replaced by a wider import. Zero assertions removed, loosened, or re-valued |
| `test_evidence_config.StepScreenshotDedupTest` / `test_step_attributed_screens` untouched | **verified** — not in the diff; ran them | 49 tests, OK |
| PR 192 interaction | **verified** — merged 192 into 193 and ran the suite | one trivial CHANGELOG conflict; 793 tests OK after resolving |

### The six-row table — all rows hold

The most important thing to check, since the whole "stop at dedup extraction" argument
rests on it. Each row is real public behaviour that existing tests pin.

1. **Filenames** — `publish`: `collector.py:156` `f"step-{self.index:02d}-{_sanitize(self.label[:MAX_LABEL_IN_FILENAME])}"`, index 0-based (`collector.py:140-141`), cap 40 (`collector.py:85`), default extension `png` (`collector.py:451`). `render_artifacts`: `evidence.py:191-195` `enumerate(steps, start=1)` → `f"{index:02d}-{safe_name}"`, default extension `svg` (`evidence.py:239-243`). **Differ.**
2. **Step text** — `collector.py:468` writes `step.screen` verbatim; `evidence.py:194` writes `step.screen + "\n"`. **Differ.**
3. **Manifest** — `evidence.json` (`collector.py:82`) with `same_as: {index, label}` and `screenshot` as a *full path* (`collector.py:511`), vs `steps-manifest.json` with `{step, screenshot, unchanged_from_previous}` and `screenshot` as a *bare name* (`evidence.py:224-231`). **Differ.**
4. **Dedup default** — `config.py:202` `dedup_step_screenshots: bool = False` gates `render_artifacts`; `publish` has no gate at all. **Differ.**
5. **Renderer protocol** — `evidence.py:161-165` falls back to text-only `render(text, path, cols, rows)`; `publish` calls `render_attributed` unconditionally (`collector.py:497`). Checked empirically, not just by reading: a text-only renderer through `render_artifacts` produces `01-one.svg` normally, while the same renderer through `publish` yields `error="'TextOnly' object has no attribute 'render_attributed'"` and `screenshot=None`. **Differ.**
6. **Render failure** — `_render_step_screens` has no `try`, so it propagates; confirmed empirically (`RuntimeError: render exploded` escaped the call). `publish` catches at `collector.py:503-507` and records `entry.error`. **Differ.**

Plus the seventh point in the body: `EvidencePublisher.identity` is a required
`RunIdentity` (`collector.py:247`) and `_render_step_screens` has no recipe/renderer/run-id
in its signature. Also true.

**Conclusion: the fuller consolidation was *not* declined on a false premise.** Routing
`_render_step_screens` through `publish()` really would need six behaviour switches on
the collector, each one a chance to move a document a consumer reads. The author's call
is defensible and the extraction they did do is the genuinely shared part — after this
PR, `render_fingerprint()` has exactly **one** production caller in the whole Python
package (`dedup.py:91`), down from two, and there is no third copy hiding anywhere.

### `SharedDedupTest` is not decorative — proved

I re-grew a *correct* local dedup three ways and ran `SharedDedupTest` each time:

- **A** — reverted `_render_step_screens` to `main`'s local fingerprint comparison, leaving the `Deduper` import in place → `test_render_artifacts_reports_the_deduper_verdict` FAILS (`stub.checked` is `[]`).
- **B** — the sneaky one: still call `deduper.check(...)`, then override the answer with a correct local copy → FAILS on `assertEqual([], renderer.rendered)`, because the local copy renders both distinct screens while the stub said reuse.
- **C** — same trick in `collector.publish` → **two** tests fail (`test_the_collector_dedupes_only_when_the_deduper_says_so` and `test_a_failed_render_forgets_through_the_deduper`).

The `_AlwaysReuses` stub is the good idea here: it returns a verdict *nothing in the
package could reach on its own* (reuse, for two demonstrably different screens), so a
module that reports it can only be taking the deduper's word. That is a delegation test
that survives a legitimate refactor of the rule itself.

`test_neither_module_has_a_deduper_of_its_own` (`assertIs(Deduper, collector_module.Deduper)`)
is the weak one — it only fails if the import is deleted — but the three behavioural
tests carry the weight and they do bite.

### Independent behavioural equivalence

I did not rely on the corpus, which exercises one scenario. I wrote a 22-case battery
covering: identical/differing/colour-only-differing screens; `dedup` on and off;
no renderer; text-only renderer; render failure on the first step, mid-run, and on
every step; failure followed by an identical screen; A-B-A sequences (the one-step
lookback boundary); labels over the 40-char cap; an uploader present and absent; and
the empty-step-list early return. Dumped `evidence.json`, `steps-manifest.json`, every
file written, and every renderer call, then ran it against `main` in a second worktree
and against the PR head.

```
$ diff -u /tmp/battery-main.txt /tmp/battery-pr193.txt
IDENTICAL: main and PR193 produce byte-identical output across all 22 battery cases
```

Sanity: the battery is not vacuous — dedup fires 6 times, the error path 4 times, and
the `render_artifacts` failure case does propagate.

### The removed dead code — the reasoning holds

`previous_fingerprint = None` in the no-renderer branch was genuinely inert, and for a
simpler reason than the PR body leads with: the loop body ended with an unconditional
`previous_fingerprint = fingerprint` two lines later, so the assignment could never be
observed. The body's *other* stated reason is also true — when `publisher.renderer is
None` nothing ever assigns `last_rendered`, so the reuse branch's `and last_rendered is
not None` guard makes the whole comparison unreachable on that path anyway. Both
readings are correct; the battery's `no-renderer-identical` case confirms the outcome
is unchanged.

## Findings

### 1. minor — the PR body's Rust-CI claim is false for this PR
**location:** PR body, "Not done" section: *"CI runs the Rust half."*

`CI (Rust)` is path-filtered to `rust/**`, `spec/**`, `conformance/**`, and its own
workflow file (`.github/workflows/rust-ci.yml:4-9`). PR 193 touches only `CHANGELOG.md`
and `python/**`. `gh-axi run list --branch fm/tp-fb-collector-a3` returns exactly two
workflows — `CI (Python)` and `Docker Image (Python)`. Compare the sibling branch
`fm/tp-fb-frames-a1`, which triggered six workflows including `CI (Rust)`.

Skipping Rust CI is *correct* hygiene — the PR changes no Rust. The problem is that the
author used "CI runs the Rust half" to discharge their own inability to run it locally,
and that discharge is void. As it happens the gap is now closed by hand: the SIGKILL
issue did not reproduce in my worktree, and `cargo test --test differential_evidence_manifest`,
`cargo test --lib evidence::dedup` and `cargo test --doc dedup` all pass at this head.

**What I'd do:** correct the sentence in the PR body. No code change.

### 2. minor — a new importable public name filed under "Changed", not "Added"
**location:** `CHANGELOG.md:44` (`### Python — Changed`), `python/termproof/dedup.py`

`termproof/dedup.py` is a new top-level module in a package whose top-level modules
*are* how consumers reach the API — `termproof.collector` is not re-exported from
`__init__.py` either, and PR #189 nonetheless treated it as public surface for an
imperative consumer. `dedup.py` declares `__all__ = ["Deduper"]`, carries a
consumer-facing docstring with a usage example, and documents `check`/`forget`/
`last_rendered` as a contract. Verified it ships:

```
$ uv build --wheel && unzip -l termproof-0.4.0-py3-none-any.whl | grep dedup
     4685  02-02-2020 00:00   termproof/dedup.py
$ uv run python -c "import termproof.dedup; print(termproof.dedup.__all__)"
['Deduper']
```

So `from termproof.dedup import Deduper` works for any consumer today, and the project
will have to keep it working. The CHANGELOG describes it in prose under *Changed* and
never announces it as new API.

Mitigating, and worth weighing: the Rust counterpart is fully public
(`pub mod dedup` at `evidence/mod.rs:25`, `pub struct Deduper` at `dedup.rs:43`), so the
Python surface is *symmetric* with Rust rather than novel. That is a real argument for
leaving it public.

**What I'd do:** this is the one thing I'd want the maintainer to choose rather than
decide for them — declare it under `Python — Added`, or make it `_dedup.py`, or leave
it and say so explicitly. See "Decision for the maintainer".

### 3. nit — two nits, one line each
- `Deduper.last_rendered` (`dedup.py:108-111`) has **no production caller in either implementation** — `grep` finds it only in `test_dedup.py` and `dedup.rs`'s own tests. This PR adds a new unused public accessor to Python to mirror an unused one in Rust, in a project that spent 0.4.0 (`d67ee92`) closing public-API doors. Faithful mirroring is a fair defence; I'd still rather it were not there.
- `termproof.dedup` is a genuinely stdlib-only module (imports `dataclasses` + `.attributed`) with a paired test file, and is a free addition to `STDLIB_ONLY` in `python/scripts/run_stdlib_tests.py:34-43`. Not added. Pre-existing pattern, though — `termproof.collector` is not in that list either.

## Merge order with PR 192

**They conflict textually, trivially, in one file, and not at all semantically.**

- `CHANGELOG.md` — both replace the single line `Nothing yet.` under `## [Unreleased]`. Real conflict; resolution is "keep both sections". Verified: `git merge origin/pr192` from the 193 head produces exactly one conflict, in `CHANGELOG.md`.
- `python/tests/test_collector.py` — both touch it, **no conflict**. 192 adds three lines inside `CaptureTest` around line 127; 193 adds imports at the top and a new `SharedDedupTest` class around line 403. Git auto-merges cleanly.
- Semantically independent: 192 only asserts that `capture_text`'s `kind` defaults to `CaptureKind.CHECKPOINT`; 193's new tests use `capture_text` but change nothing about it.

Verified end to end: merged 192 into 193, resolved the CHANGELOG by keeping both
sections, ran the full suite — 793 tests, OK.

**Order does not matter.** Whichever lands second rebases past a one-line CHANGELOG
conflict. If forced to pick: land **192 first** — it is 40 lines of docs plus one
assertion, so it clears fastest and leaves 193 as the only thing in flight touching
`collector.py`.

## What is good

- **The divergence table is honest.** This was the thing most likely to be a
  rationalisation, and it is not one. Six rows, six real differences, each pinned by
  an existing test. I checked them because I expected at least one to collapse.
- **`_AlwaysReuses` is a genuinely good test design.** Asserting a verdict that nothing
  in the package could reach on its own is what makes the delegation test survive a
  refactor of the rule. Most "did you call the collaborator" tests are mock-shaped and
  break on any legitimate change; this one is not and does not. It caught all three of
  my re-inlining attempts, including the one that kept calling the deduper.
- **The refactored `publish` now structurally mirrors the Rust one**, comment text
  included (compare `collector.py:453-518` with `rust/.../evidence/collector.rs:427-500`).
  For a project whose whole thesis is two implementations of one document format, that
  convergence is worth more than the ~8 lines of duplication removed.
- **`test_dedup.py` transcribes `dedup.rs`'s test module exactly** — 9 tests, case for
  case, same names, same comments. I diffed them by hand. That is the right way to keep
  a mirrored rule mirrored.
- **The dedup-off hard-skip is deliberately tested** (`test_dedup_off_never_asks_the_deduper`).
  Consulting a stateful object on a path where nothing reads the answer is exactly the
  kind of thing that rots into a bug later; pinning the skip was the right instinct.
- The `dedup.py` docstring example is correct — I extracted and ran it. (Nothing in
  Python CI runs it, unlike the Rust doctest, which `cargo test --doc` does execute.)

## Outside this PR

**Pre-existing, does not affect the verdict, but it changes how much "CI is green" is
worth on any change to the evidence path.**

The Python side of the conformance corpus is not enforced anywhere. `conformance/
probe_evidence_manifest.py` is run by hand only; grep finds no CI job that invokes it,
and `conformance/**` appears in the workflows solely as a *path trigger*. The only
automated reader of `conformance/corpus/evidence_manifest.expected.json` is the Rust
`differential_evidence_manifest` test, which compares **Rust output against the corpus**
and structurally cannot see Python.

I proved the hole with a realistic regression — a deduped step silently losing its
uploaded screenshot URL:

```python
# python/termproof/collector.py:489-493, reuse branch
entry.same_as = source
entry.screenshot = image
# entry.url = url      <-- removed
```

Result: **all 793 Python tests pass**, the **Rust differential test passes**, and the
corpus **no longer reproduces**. A real behaviour change to the published
`evidence.json`, fully green in CI.

For contrast, a different injected drift (adding a trailing newline to `publish`'s step
text) *was* caught by a Python unit test — so the coverage is patchy rather than absent.
But the cross-implementation document guarantee specifically has no Python-side gate.

**Recommendation for a separate change:** add a Python CI step that runs
`probe_evidence_manifest.py` and diffs it against the committed corpus, and add
`conformance/**` to `python-ci.yml`'s path filter so a corpus edit re-runs it. That is
~5 lines and it converts the PR body's strongest safety claim from a manual check into
a gate. I did not write it — out of scope for a report-only review.

Also noted, not acted on: this PR's `publish` has **no production caller inside the
package** — `grep` finds only `conformance/probe_evidence_manifest.py` and tests. It is
public API for external consumers only, which is precisely why the corpus gate above
matters more than the test count suggests.

## Decision for the maintainer

One genuine choice I will not make on the maintainer's behalf, because it is a
public-API/compat call with real arguments on both sides
(`termproof-dedup-public-surface`):

Is `termproof.dedup.Deduper` intended as public API?

- **(a) Declare it.** Move the CHANGELOG entry to `Python — Added`. Matches Rust, where
  `Deduper` is `pub`. Costs a support commitment on a 115-line module.
- **(b) Make it private** (`_dedup.py`). Matches the 0.4.0 "close every door" posture.
  Breaks symmetry with the Rust module name the docstring leans on.
- **(c) Leave as-is.** Cheapest; leaves a shipped, `__all__`-declaring, documented name
  that the CHANGELOG never announced, so nobody has decided whether it is supported.

I would pick **(a)** — the Rust side already made this type public, and a mirrored
package that hides its mirror is the more confusing of the two states — but the
compat commitment is the maintainer's to make.

I could **not** run the completion gate myself. `fm-decision-hold.sh hold` refuses with
`compatible tasks-axi is required`: this environment has `tasks-axi 0.1.2` and
`bin/fm-tasks-axi-lib.sh:37` sets `FM_TASKS_AXI_MIN=0.2.4`. That is a tool-version
mismatch in the environment, not something a report-only review should fix, so I stopped

Two other environment blocks, both consistent with prior runs: appends to

## Reproduction

Everything above, from a clean worktree at `740982d`:

```bash
git fetch origin refs/pull/193/head:refs/remotes/origin/pr193 && git checkout origin/pr193

# suite, lint, types
cd python && uv run python -m unittest discover -s tests      # 793 OK
uv run ruff check . && uv run --with mypy mypy termproof      # clean

# conformance corpus, Python side (not run by CI)
TERMPROOF_PYTHON_REPO=$PWD uv run python ../conformance/probe_evidence_manifest.py \
  | diff -u ../conformance/corpus/evidence_manifest.expected.json -   # clean

# Rust half (the author could not run this; it works here)
cd ../rust && cargo test --test differential_evidence_manifest        # 1 passed
cargo test --lib evidence::dedup && cargo test --doc dedup            # 9 + 1 passed

# merge interaction with 192
git checkout -B tmp origin/pr193 && git merge origin/pr192            # 1 conflict, CHANGELOG.md only
```

The 22-case differential battery is at `/tmp/dedup_battery.py`; run with
`TP_REPO=<repo>/python uv run python /tmp/dedup_battery.py` from `<repo>/python`
against two checkouts and diff the output.

PR #194, opened 2026-08-20
evidence: append the captured checkpoint screens to a cast as held trailing frames

md-mt:
`append_checkpoint_frames`, in both implementations, beside `cast_video`.

A cast stops at whatever the last keystroke painted, which is rarely the
evidence — the checkpoints that made the run reviewable have already scrolled
away by then. This writes them back onto the end, in capture order, each one
repainting the whole grid and held long enough to read, so a recording ends by
replaying the evidence sequence. That is the reason a reviewer watches one
artifact instead of opening fifteen stills.

## The surface

| | Rust | Python |
|---|---|---|
| Where | `evidence::cast_video::append_checkpoint_frames` | `termproof.cast_video.append_checkpoint_frames` |
| Screens | `&[CapturedStep]` | `Sequence[CapturedStep]` |
| Hold | `hold_seconds: Option<f64>` | `hold_seconds: float = DEFAULT_CHECKPOINT_HOLD` |
| Default hold | `DEFAULT_CHECKPOINT_HOLD` = 3.0 | same |
| Returns | `Result<usize, String>` | `int` |
| Bad hold | `Err` | `ValueError` |

`Option` for a Python keyword default and `Err` for a `ValueError` are the
house patterns, so the two are mirrors rather than merely siblings.

## What it does to the file

Appends only. The header and every recorded event are left exactly as the
session wrote them, and the new events carry on from the last timestamp in the
file rather than restarting at zero — still a valid asciinema v2 cast, with
timestamps that only increase.

Three decisions worth a reviewer's attention, each argued in the source:

- **A closing no-op event.** A cast ends at its final event, so the last
  checkpoint — the frame a reviewer skipped to the end for — would otherwise
  flash rather than hold. One SGR reset at `+hold` keeps the recording running
  through it.
- **Every captured step, `CaptureKind::Failure` included.** Filtering by kind
  would drop the run's failure screen, which is exactly the wrong one to drop.
- **The first checkpoint lands one hold after the session's final event**, which
  leaves the live ending up for a beat before the replay starts. One number, no
  second magic constant for the lead-in.

A run that captured nothing is a silent no-op — the file is not opened, so it
does not even have to exist. Nothing about any of it needs the session to still
be running; it reads and appends to a closed file.

## Tests

Mirrored unit tests on both sides cover capture order, timestamps continuing
from the session's end and only increasing, the hold being honoured (explicit
and defaulted), and the empty case being a silent no-op. Also: the appended
cast still parses (`read_cast` / the header and every event), and still
*replays* through the real emulator to the right screens — the stronger
statement, since a player parses every payload too.

Python: 791 pass (`unittest discover`), `ruff check` clean, `mypy` clean on the
changed module, and `scripts/run_stdlib_tests.py` green. The four replay tests
are `skipUnless(pyte)` so that stdlib-only gate keeps passing; the parse test is
not, so "still a valid cast" is asserted in every environment.

## Conformance

The evidence-manifest pair now writes a fixed cast into the publish directory
after publishing and appends the captured steps to it, so the two
implementations' appended events are compared byte-for-byte alongside the step
text — a manifest agreeing about a recording's path is not agreeing about the
recording. The scenario's seven captures include the SGR-escaped screen and the
two-line one, so the payloads exercise both JSON escaping and the carriage
returns a raw terminal needs.

That required pinning the Python event encoder to `serde_json`'s shape (compact
separators, `ensure_ascii=False`), since Python's `json` defaults to neither.
`conformance/corpus/evidence_manifest.expected.json` is regenerated by the
documented command.

## What I could not run

**The Rust tests did not execute in this environment.** Every freshly compiled
binary is `SIGKILL`ed on exec here — including a `rustc` hello-world outside the
repo — so `cargo test` never gets past process start. This is environmental, not
a property of the change. What I did verify:

- `cargo clippy -p termproof --all-targets` clean, `cargo fmt` applied,
  `cargo doc` adds no new warnings (the 6 present are pre-existing).
- The two `BASE_CAST` literals decode to identical bytes (checked
  programmatically, not by eye).
- Every appended payload in the regenerated corpus equals the repaint prefix
  plus that step's own `step-NN.txt` contents — and the Rust half already
  produces those byte-identically, which is what the existing harness asserts.
- `serde_json`'s escaper writes `\u00xx` lowercase for control bytes and leaves
  non-ASCII raw, matching what the pinned Python encoder produced
  (`test_the_event_encoder_is_pinned_to_what_rust_emits` holds that end).

So the Rust half is verified by construction and by CI, not by a local run. CI
is the real check.

## Not done

No version bump; the entry is under `[Unreleased]` per the changelog convention.

`AGENTS.md` (and its `CLAUDE.md` symlink) is new at the repo root, carrying two
notes: the mirror rule with a pointer to `conformance/README.md`, and the
`run_stdlib_tests.py` gate that a pyte-using test in a listed module silently
breaks. Both cost me a cycle here.


md-mt:
CI is green: 20 passed, 0 failed, 3 skipped.

That closes the verification gap in the description — the Rust unit tests and the extended `differential_evidence_manifest` conformance test both ran and passed on ubuntu-latest and macos-latest, so the Rust half is no longer verified only by construction. Nothing in the branch changed; only the local sandbox could not exec compiled binaries.

md-mt:
Both majors fixed in `70e1d1f`. Thank you for the repro — the scroll-region case is one I would not have found, and the mutation results on the hold and the rounding were exactly right.

## 1 — the repaint now resets the scroll region

`REPAINT_PREFIX` / `_REPAINT_PREFIX` is `\x1b[m\x1b[r\x1b[H\x1b[2J`. I reproduced your case before changing anything, and it behaves as you reported:

```
scroll region    current : ['L1','L2','L5','L6','L7','L8','','']
                 fixed   : ['L1','L2','L3','L4','L5','L6','L7','L8']
region+origin    current : ['','','L5','L6','L7','L8','','']
                 fixed   : ['L1','L2','L3','L4','L5','L6','L7','L8']
```

**I did not use a soft reset**, which was my first instinct and would have covered charsets and autowrap too. `avt` implements DECSTR (`parser.rs:817`) and `pyte` does not — no `!p` handler in its `streams.py` dispatch table. The two replay paths this project renders through would then disagree about what the frame says, which is worse than the state it would have reset. That reasoning is in the constant's doc comment so the next person does not re-derive it.

Pinned by a unit test on each side (base cast sets `\x1b[3;6r\x1b[?6h`, asserts all eight rows), and the conformance `BASE_CAST` now opens with `\x1b[3;20r` so the corpus records the prefix that undoes it.

## 2 — the default hold and the rounding

**Default hold.** Both sides now assert the literal, not the constant: `assertEqual([0.0, 3.0, 6.0, 9.0], ...)` / `assert_eq!(times, vec![0.0, 3.0, 6.0, 9.0])`, with a comment saying why it is spelled out. Your `3.0 -> 5.0` Python mutant dies on it.

**Rounding.** I went further than adding an assertion, because you were right that two algorithms is the trap. Python now transcribes Rust's expression rather than calling `round(at, 6)`:

```python
scaled = at * 1e6
floor = math.floor(scaled)
return (floor + 1 if scaled - floor >= 0.5 else floor) / 1e6
```

I swept 501,584 values — your reachable grid plus 500k random, plus the nasty ones (`5e-7`, `61.3999995`, `999.9999995`) — against Rust's `f64::round` semantics simulated exactly:

| Python expression | divergences from Rust |
|---|---|
| `round(at, 6)` (before) | 173 |
| `round(at * 1e6) / 1e6` | 108 |
| transcribed floor form (now) | **0** |

Note the middle row: scaling but leaving the tie-break to Python's `round` is *not* enough, because half-integers do occur (`5e-7 * 1e6` is exactly `0.5`, half-to-even sends it to 0, Rust sends it to 1). Both sides now assert the written timestamps as **text** — `["0.1","0.3","0.5","0.7","0.9"]` — since a parsed float cannot tell `0.3` from `0.30000000000000004`.

**Conformance.** Added `session-with-fractional-hold.cast`: the same seven steps at an explicit `0.2` from a base of `0.1`. That covers the `hold_seconds` path the default cast could not reach and lands on timestamps binary addition gets wrong, so the corpus pins the six-decimal answer across implementations. The default-hold cast stays, so `3.0` is pinned there too. Corpus regenerated by the documented command, reproduces byte-for-byte.

## 3 — the minor came along, because your fix for it was wrong and so was mine

You suggested requiring `round(hold, 6) > 0`. I implemented that, and my own test caught it: `5e-7` passes such a check and still collides frames 1 and 2, because `k * 5e-7` rounds to `1e-6, 1e-6, 2e-6`. Rounding *up* to a microsecond is not the same as advancing by one.

The floor is therefore on the unrounded hold: `MIN_CHECKPOINT_HOLD = 1e-6`, public on both sides. A true gap of at least 1e-6 survives rounding as a gap, for every multiple. `0.0, -1.0, NaN, inf, 1e-9, 5e-7` are all rejected on both sides; `1e-6` is accepted and its timestamps are asserted to increase.

## Nits

- **Error surface** (nit 4, third bullet): the Python docstring now documents `OSError` alongside `ValueError`, and "no header line" is corrected to "the cast is empty", which is what the code actually checks.
- **Exponent formatting** and **`[true,"o","x"]`**: left alone deliberately. Both need input no recorder produces, and the microsecond floor now puts the exponent case out of reach for any accepted hold. Recorded here rather than papered over.

## Verification

Python: 793 tests, `ruff` clean, `mypy` clean, stdlib gate `145 passed, 6 skipped`, corpus reproduces byte-for-byte. Rust: `clippy --all-targets` clean, `fmt --check` clean, `cargo doc` still 6 pre-existing warnings and no new ones. The two harness halves' `BASE_CAST` / `FRACTIONAL_BASE_CAST` literals are checked to decode identically by script rather than by eye.

Same caveat as before on my side: I still cannot execute compiled binaries here, so CI is where the Rust half runs. Given you have a working environment, the scroll-region test and the two new casts are the places worth a second look.

md-mt:
# PR 194 — evidence: append the captured checkpoint screens to a cast as held trailing frames

**VERDICT:** CHANGES NEEDED
**MERGE RISK:** medium — the produced cast is valid and plays everywhere I tried, but the "repaints the whole grid" claim is false whenever the recorded session left a scroll region set (any full-screen TUI). Rows of the evidence screen are then silently deleted from the appended frame. Nothing in-tree calls this yet, so merging regresses nothing; the risk is that the wrong semantics become load-bearing for the queued work built on top.
**CI:** 20 passed, 0 failed, 3 skipped (23 total) on the current head `0886a6b`. Genuinely meaningful coverage for this change: `Lint, type-check and test` on ubuntu + macos, Python unit tests on 3.11/3.12/3.13, the stdlib-only gate, `cargo semver-checks`, `cargo deny`, and the TUI evidence build. The 3 skips are `Deploy docs`/`Publish`/pages-gate jobs, not skipped tests. CI is the only place the Rust half runs for the author — see "What is good".
**STALENESS:** Clean. `git merge-base main pr194` = `origin/main` = `f033dcc`; the branch is not behind. Applies cleanly and I ran both full suites on the head.

## Summary

Adds `append_checkpoint_frames` to both implementations: it reads a closed asciinema v2 cast, appends one output event per captured step (each prefixed with a screen repaint, spaced one hold apart, timestamps continuing from the file's largest existing timestamp), then a final SGR-reset event so the last checkpoint is held rather than flashed. The file-level mechanics are right and I verified them against real asciinema tooling, not just the tests. The defect is in the repaint prefix, which resets the pen and the cursor but not the scroll region, so on exactly the class of application this project exists to record the appended frame is silently wrong. Two secondary problems: the Python default hold and the six-decimal rounding on both sides are pinned by no test at all, and rounding is the one place the two implementations use different algorithms.

## How I verified

Worktree at PR head `0886a6b`, all scratch changes reverted (`git status` clean at the end).

- Full suites on the head: Python `791 passed`; Rust `399` lib + all 16 integration binaries + 8 doctests, exit 0. **Rust binaries run in my environment**, so I executed the half the author could only reason about.
- Gates: `ruff check` clean, `mypy termproof` clean (41 files), `scripts/check_version.py` clean, `scripts/run_stdlib_tests.py` `143 passed, 5 skipped`, `cargo fmt --check` clean, `cargo clippy -p termproof --all-targets -- -D warnings` clean.
- Independent cast readers: `asciinema 2.4.0` (`asciinema cat`, the reference implementation) and `agg 1.9.0` (Rust, avt — the engine the asciinema player uses). Both parse and replay the appended cast.
- Cross-implementation byte diff over 16 hand-built cases plus a 31,360-value rounding sweep.
- Mutation testing: 14 Python mutants, 11 Rust mutants, each run against the new unit tests and (Rust) the conformance test.

## Findings

### 1. major — the repaint does not reset the scroll region, so it silently drops rows of the evidence screen

**Location:** `python/termproof/cast_video.py:80`, `rust/crates/termproof/src/evidence/cast_video.rs:76` (`_REPAINT_PREFIX` / `REPAINT_PREFIX`)

**Status: VERIFIED**, in two independent terminal engines.

The prefix is `\x1b[m\x1b[H\x1b[2J` — SGR reset, cursor home, erase display. `\x1b[2J` erases the display but leaves DECSTBM (the scroll region) exactly as the live session set it, and nothing resets DECOM (origin mode) either. A full-screen TUI — vim, less, htop, tmux, most curses apps — sets a scroll region and does not clear it on exit.

Concrete failure, an 8-row cast whose only live event is the 10 bytes `\x1b[3;6r`:

```
captured screen : ['L1','L2','L3','L4','L5','L6','L7','L8']
replayed frame  : ['L1','L2','L5','L6','L7','L8','','']     <- L3 and L4 gone
```

The 8 rows are written into a 4-row scroll region, so the region scrolls and two rows of evidence are destroyed. With origin mode also set (`\x1b[3;6r\x1b[?6h`) four of eight rows are lost: `['','','L5','L6','L7','L8','','']`.

Reproduced through:
- the project's own `frames_from_cast` (pyte), and
- `agg 1.9.0` rendering the same cast to a GIF — the rendered frame shows `L1 L2 L5 L6 L7 L8` where the fixed version shows all eight.

This is the worst failure shape for an evidence artifact: the cast is still valid, still parses, still plays, and the screen is simply not the screen that was captured. Nothing warns. It also directly contradicts the PR body, the changelog entry and both docstrings ("each one repainting the whole grid", "a complete picture, not a delta").

**What I would do.** Add `\x1b[r` (DECSTBM reset, no parameters) to the prefix, before the cursor-home: `\x1b[m\x1b[r\x1b[H\x1b[2J`. I verified that exact string restores all eight rows in both pyte and avt, for both the DECSTBM and the DECSTBM+DECOM case. (`\x1b[?6l` is not additionally needed — origin mode only relocates the origin relative to the margins, so resetting the margins is sufficient — but it is harmless if wanted for explicitness.) Then pin it: a unit test on each side whose base cast contains a scroll region, and a scroll region added to the conformance `BASE_CAST` (`conformance/probe_evidence_manifest.py:64`, `rust/crates/termproof/tests/differential_evidence_manifest.rs:63`) so the corpus holds both implementations to it.

Not a blocker only because nothing in-tree calls `append_checkpoint_frames` (verified by grep: the sole callers are the conformance probe and the differential test), so merging regresses no existing behaviour.

### 2. major — the Python default hold and the rounding on both sides are pinned by nothing; the differential harness pins Rust only

**Location:** `python/termproof/cast_video.py:75` and `:225`; `rust/.../cast_video.rs:70` and `:435`; tests at `python/tests/test_cast_video.py:380`, `rust/.../cast_video.rs:780`

**Status: VERIFIED** by mutation.

Changing `DEFAULT_CHECKPOINT_HOLD` from `3.0` to `5.0` **in Python only** leaves the entire 791-test Python suite green *and* the Rust conformance test green. The two defaults can silently drift apart. The PR body's "Default hold … same" is true today and enforced in one direction only.

The mechanism: every hold assertion on both sides is written in terms of the constant —

```python
hold = DEFAULT_CHECKPOINT_HOLD
self.assertEqual([0.0, hold, 2 * hold, 3 * hold], [at for at, _ in self._events()])
```

— which constrains the *spacing structure* (correctly, and it killed my "no lead-in" and "drop the terminator" mutants) but never the *value*. The only artefact that pins `3.0` is `conformance/corpus/evidence_manifest.expected.json`, and that corpus is read only by `differential_evidence_manifest.rs:73`. Nothing in CI regenerates or verifies the Python probe's output — `conformance/README.md` says regenerate deliberately. Mutation confirms the asymmetry exactly: Rust's `3.0 -> 5.0` is `unit=SURVIVED, conformance=KILLED`; Python's is survived by everything.

Worse, the six-decimal rounding survives **every** test on both sides — it was the only mutant to survive both the Rust unit tests and the conformance test. And rounding is precisely the one place the two implementations are not the same code:

| | Python `:225` | Rust `:435` |
|---|---|---|
| | `round(at, 6)` | `(at * 1e6).round() / 1e6` |
| semantics | exact-decimal, half-to-even | scale, half-away-from-zero, divide |

Untested plus algorithmically different is the classic drift setup.

**Mitigating, and I checked it rather than assuming:** across 31,360 *reachable* timestamps (140 session-end times with ≤6 decimals × 14 plausible holds × 16 frames) the two produce **byte-identical** output. I could not construct a reachable divergence. Divergence needs a base timestamp with 7 decimals (e.g. `61.3999995` → Rust `61.4`, Python `61.399999`) or magnitudes ≥ 1e12 s. So this is a latent trap, not a live bug — which is why it is major and not a blocker.

**What I would do.** One assertion per side pinning the literal `3.0`, and one pinning a timestamp that actually requires rounding (same literal string asserted on both sides). Better still, put a fractional hold into the conformance scenario so the corpus pins the rounding across implementations rather than leaving it to two independently-written tests.

### 3. minor — a positive, finite hold can still stall the timestamps, and the test named for that case does not cover it

**Location:** `python/termproof/cast_video.py:268`, `rust/.../cast_video.rs:370`; tests at `python/tests/test_cast_video.py:465`, `rust/.../cast_video.rs:893`

**Status: VERIFIED.**

`hold_seconds=1e-9` passes validation (finite, positive) and the six-decimal rounding then collapses every appended timestamp to the same value. Both implementations produce, byte-identically:

```
[0.0,"o","a"]
[0.0,"o","[m[H[2Ja"]
[0.0,"o","[m[H[2Jb"]
[0.0,"o","[m"]
```

Anything below ~5e-7 does this. The rejection message and both docstrings justify the check by saying such a hold "would stall or reverse the timestamps this promises to keep increasing" — but the hold that actually stalls them is accepted. `a_hold_that_would_stall_the_timestamps_is_rejected` / `test_a_hold_that_would_stall_the_timestamps_is_rejected` covers `0.0, -1.0, NaN, inf` only. The assertion is not tautological (it kills the "drop the validation" mutant on both sides), but it does not test the case it is named after.

**What I would do.** Reject a hold whose rounded delta is zero — e.g. require `round(hold, 6) > 0` — and add that value to the existing loop of bad holds on both sides.

### 4. nit — three degenerate-input divergences between the two implementations, bounded and unreachable from a recorder

**Status: VERIFIED**, recorded so the parity claim is precisely bounded rather than left open. None of these need action for merge.

- **Exponent formatting.** `serde_json`/ryu writes `1e-6`, Python's `json` writes `1e-06`, for the same double. Reachable only when a timestamp's shortest repr uses exponent notation, i.e. `hold < 1e-4` on a cast starting near zero, or ≥ 1e16 s. The default hold can never reach it.
- **`[true,"o","x"]` as an event line.** `isinstance(True, (int, float))` is `True` in Python, so `_last_event_time` reads it as `1.0`; Rust's `as_f64()` rejects it. Same malformed input, appended frames one second apart between the two.
- **Error surface.** Missing file with non-empty steps: Rust `Err("No such file or directory")`, Python `FileNotFoundError`. That matches the house `Result`↔exception pattern, but the Python docstring documents only `ValueError`. Relatedly, the documented "raises if the cast has no header line" is really "raises if the file is empty" — a plain text file with any first line is appended to without complaint, identically on both sides.

## What is good

Several things here are subtle and the author got them right; I checked each rather than taking the PR body's word.

- **The closing terminator event is a real insight, and it works.** Rendered through `agg`, the GIF frame durations are `[100, 1150, 1250, 3000, 3000, 3000, 3000] ms` — the final `ERROR` checkpoint is held a full 3000 ms because the terminator event follows it. Without it that frame would be the cast's last event and would flash. Dropping it kills 6 Python tests.
- **The cast really is valid, and in more than one reader.** `asciinema cat` (the reference implementation) and `agg 1.9.0` (avt, the player's engine) both parse header, event shapes and every payload. The header is left byte-identical including its `env`/`timestamp` fields. Timestamps continue from the session (`2.5 → 5.5 → 8.5 → 11.5`) rather than restarting at zero — confirmed both in the file and in the renderer's cumulative frame times.
- **Leftover SGR between frames is handled.** The leading `\x1b[m` does its job: in the rendered GIF the live session's bold-green `DASHBOARD READY` paints plain in the following checkpoint frame. Cursor position is handled too (`\x1b[H`). It is only the scroll region that is missed.
- **The empty case is a true no-op, not a same-bytes rewrite.** Verified two ways: by reading (`if not steps: return 0` is the first statement, before any file access) and by mutation — inserting a read-then-write into that branch fails `test_no_checkpoints_is_a_silent_no_op`, because the test also calls it on a path that does not exist.
- **The missing-final-newline guard is real and correctly scoped.** I also checked the harder case: a cast truncated *mid-event* was already unparseable before the call and remains so afterwards — the function neither fixes nor worsens it. That is honest scoping, not an oversight.
- **The `\r\n` conversion is necessary, not defensive.** Removing it reproduces the staircase and kills the test on both sides.
- **The test suites are strong.** 11 of 14 Python mutants killed by the new test class alone (reverse order, delta-instead-of-repaint, restart-at-zero, no lead-in, drop terminator, drop newline guard, drop validation, drop CRLF, unpin encoder, filter failures). 10 of 11 Rust mutants killed by unit tests, and the one unit-survivor is caught by conformance. The three survivors are finding 2 above.
- **The conformance pair was extended, and honestly.** The corpus regenerates **byte-for-byte** from the documented command (`TERMPROOF_PYTHON_REPO=$PWD python ../conformance/probe_evidence_manifest.py`) — it was not hand-edited. It covers capture order, timestamp continuation, the default hold, JSON escaping of SGR and non-ASCII, and the CRLF conversion. It does not cover an explicit `hold_seconds`, the error paths, the empty case, or the missing-newline case; those are unit-tested per-implementation only.
- **I ran the Rust half the author could not.** The PR's "What I could not run" section is candid and its reasoning was sound; for the record, on this machine the Rust suite is fully green (399 lib tests, all integration tests, clippy `-D warnings` clean, fmt clean), so the by-construction argument holds up under execution.
- Conventions match: `&str` for paths mirrors `convert`/`output_path_for` in the same Rust module; `__all__` updated; changelog under `[Unreleased]` with no version bump; no doc surface lists per-function APIs, so nothing was missed there.

## Outside this PR

- **The differential harness is one-directional by design.** The corpus pins Rust to a frozen snapshot of Python; nothing in CI regenerates or verifies the Python side against it, so any Python-side behaviour drift ships silently until someone regenerates deliberately. This is pre-existing (`conformance/README.md`: "Only regenerate deliberately") and predates this PR. It is listed as finding 2 only in so far as *this* PR's central parity claim leans on it while introducing a surface where the two sides use different algorithms. The general fix — a CI job that regenerates the corpus and fails on drift — is a separate change.
- `AGENTS.md` (34 lines) and the `CLAUDE.md` symlink are new at the repo root and unrelated to the feature. Both are well-formed (`CLAUDE.md` is a proper mode-120000 git symlink) and the author flags them in the PR body. Not worth blocking on; noting only that they widen the diff.

## What I could not verify

Nothing material. Every claim above was executed rather than reasoned about, with two exceptions I want to be explicit about:

- I could not test against the browser-based asciinema player itself, only against `agg`, which embeds the same `avt` parser. `asciinema cat` covers the reference Python parser independently.
- The scroll-region failure is demonstrated with a synthetic base cast. I did not find a scroll region in this repo's own `python/examples/` TUIs, so the repo's current recipes would not hit it — the exposure is real applications recorded by downstream users, not the in-tree examples.

md-mt:
# PR 194 — evidence: append the captured checkpoint screens to a cast as held trailing frames

*Second review, of the fix round `0886a6b..70e1d1f` only. The first review's verdict was
CHANGES NEEDED on two majors, one minor and one nit.*

**VERDICT:** MERGE WITH NITS
**MERGE RISK:** low — the scroll-region defect is genuinely gone and the two implementations now
agree on every timestamp I could reach. What is left is test coverage, not behaviour: the
rounding rule the fix round transcribed is pinned by nothing, and three comments added in this
round state as fact things that are not true.
**CI:** 20 passed, 0 failed, 3 skipped (23 total) on `70e1d1f`. The 3 skips are deploy/publish
gates, not tests. Meaningful for this change.
**STALENESS:** The branch is now two commits behind the default branch (base is still the old
head). Merging conflicts, but only in `CHANGELOG.md`, where both sides added entries under
`[Unreleased]`. **Verified:** I resolved it by keeping both blocks and ran the merged tree —
807 Python tests pass, the conformance corpus still reproduces byte-for-byte, and the
differential test passes. No semantic staleness.

## Summary

Both majors are addressed and I confirmed the important one by construction and by execution
rather than by reading: the repaint now resets the scroll region, and the regression tests on
both sides fail when the fix alone is reverted. The rounding is now one algorithm transcribed
into two languages, and a 27,306-case cross-implementation sweep through both real
implementations found zero divergences in value on any input a recorder produces. The
remaining problems are all in the guards, not the code: the transcribed rounding survives being
swapped back for `round(at, 6)` with every test green and the corpus regenerating identically,
the microsecond floor survives being swapped back for the first review's weaker suggestion with
every Python test green, and the comments added around both claim the opposite.

## How I verified

Worktree at `70e1d1f`; every scratch change reverted (`git status` clean at the end). I can run
compiled binaries, so the Rust half was executed rather than reasoned about.

- Full suites at the head: Python `793 passed`; Rust `545 passed, 0 failed` across lib,
  all integration binaries and doctests.
- Gates: `ruff` clean, `mypy` clean (41 files), version-drift check clean, the stdlib-only
  gate `145 passed`, `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings`
  clean, `cargo semver-checks` — 196 checks, 196 pass, "no semver update required" (the two new
  public constants are additive).
- The conformance corpus regenerates byte-for-byte from the documented command.
- Mutation: seven targeted mutants, listed per finding below.
- Cross-implementation sweeps: 384 + 27,306 (base, hold) pairs driven through *both real
  implementations* end to end — cast written, appended, timestamps read back as text — plus
  a 40,000-value float-parse comparison.
- Two independent terminal engines for the scroll-region case: `pyte` (the Python replay path)
  and `avt` 0.16 (the Rust replay path, and the engine the player and `agg` use). Both casts
  also render through `agg` 1.9.0 without complaint.

Not re-litigated, per the first review, which cleared them: cast validity against real
tooling, header preservation, the closing terminator, the CRLF conversion, the empty-steps
no-op, the missing-final-newline guard. Nothing in the delta touches any of them.

## Findings

### 1. minor — the transcribed rounding is pinned by nothing, and the new tests say it is

**Location:** `python/termproof/cast_video.py:241`, `rust/crates/termproof/src/evidence/cast_video.rs:477`;
tests at `python/tests/test_cast_video.py:396` and `rust/.../cast_video.rs:859`

**Status: VERIFIED** by mutation, three ways.

Replacing the whole of `_round_micros` with `return round(at, 6)` — i.e. undoing exactly the
change this round made — leaves:

- the Python suite at **793 passed, OK**;
- the conformance corpus regenerating **byte-for-byte identical** (`diff` empty), so the
  differential test cannot see it either.

The reason is that the new fractional case does not discriminate. `0.1 + k * 0.2` rounds the
same way under both rules, so `["0.1","0.3","0.5","0.7","0.9"]` is what *either* rule writes.
The same is true of the new `session-with-fractional-hold.cast` in the corpus.

Both new tests carry a comment asserting the opposite:

```
# ... and `round(at, 6)` here would not have produced it -- it
# breaks halves to even where Rust breaks them away from zero.
```

It would have produced exactly it. `conformance/README.md` leans the same way. This is the
trap: the next person reading these three places will believe the drift guard exists.

**Two rules really do differ on accepted input, so the guard is worth having.** Verified
through both implementations, base `1.0`, `hold_seconds=1.5e-6` (legal — above the new floor):

```
today, both sides : 1.000002, 1.000003, 1.000005, 1.000006, 1.000007
with round(at, 6) : 1.000001, 1.000003, 1.000004, 1.000006, 1.000007
                    ^^^^^^^^             ^^^^^^^^
```

**What I would do.** Add that case — a base of `1.0` and `hold_seconds=1.5e-6`, asserted as
text on both sides — to the existing fractional test. It kills the mutant. Then correct the
three comments. (`base 0.0, hold 3.5e-6` works too: `4e-6` versus `3e-6` on the first frame.)
If it is worth pinning across implementations rather than twice independently, the same hold
in the conformance scenario does it.

### 2. minor — the microsecond floor is not pinned on the Python side, and the value chosen for the bad-hold list is the one value that cannot pin it

**Location:** `python/termproof/cast_video.py:314`, `rust/.../cast_video.rs:401`; tests at
`python/tests/test_cast_video.py:506`, `rust/.../cast_video.rs:1002`

**Status: VERIFIED** by mutation.

Swapping the Python check back to the first review's suggestion —
`round(hold_seconds, 6) <= 0` — leaves the Python suite at **793 passed, OK**. The new
`5e-7` entry in the bad-hold list is what should have caught it and cannot, for a reason
worth spelling out (see the next section): `round(5e-7, 6)` is `0.0`, so the weaker check
rejects `5e-7` too.

The Rust side does not have this hole — the equivalent weakening (`round_micros(hold) <= 0.0`)
is killed by its bad-hold list, because Rust's rule sends `5e-7` to `1e-6` where Python's sends
it to `0.0`. So the same test, written to look the same on both sides, pins the check on one
side only.

Separately, and in the same shape as the first review's finding 2: **`MIN_CHECKPOINT_HOLD`
itself can drift between the implementations.** Changing it to `1e-3` on the Rust side leaves
the Rust suite green; changing it to `1e-3` on the Python side leaves the Python cast-video
tests green. Both accept-case assertions are written in terms of the constant, so they follow
it wherever it goes. A hold of `1e-4` would then be accepted by one implementation and
rejected by the other.

**What I would do.** Put `6e-7` (or `7.5e-7`) in the bad-hold list on both sides — verified
below to be accepted by the weaker check and to collide — and assert the literal `1e-6` for
the constant on both sides, exactly as was just done for `DEFAULT_CHECKPOINT_HOLD`.

### 3. On the disagreement about the first review's suggestion — the conclusion is right, the counterexample is not

Asked directly, in both directions, so here it is in both directions.

**The suggestion was wrong. VERIFIED.** `round(hold, 6) > 0` is not sufficient. Driven through
the real Python implementation with that check substituted, on a cast ending at `0.0`:

```
hold=6e-07     ACCEPTED -> 0.0, 1e-06, 1e-06, 2e-06, 2e-06     <- frames 1 and 2 collide
hold=7.5e-07   ACCEPTED -> 0.0, 1e-06, 2e-06, 2e-06, 3e-06     <- frames 2 and 3 collide
```

Rounding *up* to a microsecond is not advancing by one, which is the author's point, and it
holds. The floor on the unrounded hold is the right shape of fix, and `1e-6` is the right
value.

**The stated counterexample is wrong. VERIFIED.** `5e-7` does not pass that check:

```
>>> round(5e-7, 6)
0.0
```

The double nearest `5e-7` is `4.9999999999999997737...e-07` — just *below* the exact half — so
`round` takes it down to zero and the suggested check rejects it. What is true is the other
half of the sentence: `5e-7` *would* collide frames 1 and 2 if it were accepted, because the
scaled value is exactly `0.5` and the away-from-zero rule sends both `5e-7` and `1e-6` to
`1e-6`. So the test comment's claim about collision is right and its claim about the check is
not — and that is why the case was picked that cannot pin the check (finding 2).

None of this changes what shipped. It changes one comment and one test value.

### 4. nit — the guarantee in the new constant's doc comment does not quite hold at the floor

**Location:** `python/termproof/cast_video.py:106`, `rust/.../cast_video.rs:106`

**Status: VERIFIED**, in both implementations.

> a gap of at least 1e-6 in the input is still a gap after rounding

Not always. With `hold_seconds` exactly `1e-6` and a cast whose last timestamp carries an exact
half-microsecond, two appended frames land on the same timestamp — the failure the constant
exists to prevent:

```
base 999.9999995, hold 1e-6
  -> 1000.000001, 1000.000001, 1000.000002, 1000.000003, ...
```

Bounded, and I measured the bound: **0 collisions in 300,000 random six-decimal base
timestamps** at `hold = 1e-6`, and **0 collisions at any hold above the floor**, including
`1.000001e-6`. It needs the floor exactly and a half-microsecond base together. The new test
asserts the increase only from a base of `0.0`, which is the case that cannot fail.

**What I would do.** Nothing to the code. Soften the sentence to say what is true — the floor
is what makes the increase hold for every timestamp a recorder writes — or, if the absolute
claim is wanted, compare against the previous rounded value rather than trusting the input gap.

### 5. nit — the reason given for leaving the exponent divergence alone is not correct

**Status: VERIFIED.**

> the microsecond floor now puts the exponent case out of reach for any accepted hold

It does not. The floor is itself an accepted hold, and it is one of the cases:

```
base 0.0, hold = MIN_CHECKPOINT_HOLD (1e-6)
  Rust   : 1e-6,   2e-6,   3e-6,   4e-6,   ...
  Python : 1e-06,  2e-06,  3e-06,  4e-06,  ...
```

Every appended timestamp below `1e-4` diverges in bytes — sometimes as `1e-6` versus `1e-06`,
sometimes as `0.000011` versus `1.1e-05`, since the two switch into exponent notation at
different magnitudes. In my 27,306-case sweep this accounted for 24 divergences and in the
smaller one 12, all of them the same numeric value written differently.

The decision to leave it is still defensible — no recorder produces a session that ends near
zero *and* a caller asking for a sub-100-microsecond hold. Only the reason needs correcting,
since it is the reason recorded on the PR for closing the item.

### 6. nit — `[true,"o","x"]` is genuinely out of reach, as the author says

**Status: VERIFIED** (behaviour unchanged by this round; `_last_event_time` is untouched in the
delta). Python still reads the boolean as `1.0`, Rust still rejects it. A boolean where the
format requires a number is not something any recorder emits. Agreed: leave it.

## What is good

- **The scroll-region fix is real, and I did not take the code's word for it.** With `\x1b[r`
  removed from the prefix and nothing else changed, the new test fails on both sides
  (`test_a_scroll_region_left_by_the_session_does_not_eat_rows` in Python,
  `a_scroll_region_left_by_the_session_does_not_eat_rows` in Rust) and the differential
  conformance test fails as well, because the corpus now records the prefix bytes. Restored, the
  eight-row evidence screen comes back whole in `pyte`, in `avt`, and in the sequence
  `agg` renders. The DECSTBM+origin-mode case is covered too — the harder of the two.
- **The reason given for not using a soft reset is accurate, and load-bearing.** I checked it:
  `pyte` ignores `\x1b[!p` (margins unchanged after feeding it) and honours `\x1b[r` (margins
  reset). A soft reset really would have made the two replay paths disagree. That reasoning
  belongs in the comment and it is right.
- **The default-hold pin works on both sides.** `3.0 -> 5.0` in Python now fails
  `test_the_hold_is_the_gap_between_frames_and_defaults`; the same mutation in Rust now fails
  its unit test rather than surviving to be caught only by conformance. That is exactly the
  asymmetry the first review reported, closed.
- **One rounding algorithm instead of two is the right call, and parity holds.** 27,306
  (base, hold) pairs driven through both implementations end to end, on top of the earlier 384:
  **zero divergences in value** for six-decimal base timestamps (8,880 cases) and for
  full-precision doubles (8,880 cases). The only value divergences, 401, all come from
  half-microsecond bases, and all 401 are explained by something else entirely — see below —
  not by the rounding.
- **Putting the scroll region into the conformance base cast was the right instinct.** It means
  the corpus, not just two hand-written tests, holds both implementations to undoing it.
- **The corpus was regenerated, not edited.** It reproduces byte-for-byte from the documented
  command, at the head and after merging the current default branch.
- **The docstring corrections are accurate.** "the cast is empty" is what the code checks, and
  `OSError` is what Python raises.
- **The candour is worth keeping.** Every claim in the fix comment is checkable, and three of
  them turned out to be checkable and wrong — which is far more useful than vagueness.

## Outside this PR

- **The JSON parser this repository reads timestamps with is one ULP off on about 5% of
  full-precision floats, and that is the real parity hole.** All 401 value divergences in my
  sweep trace to it: `serde_json` 1.0.151 and Python's `json` parse the same literal to
  different doubles. Demonstrated without leaving Rust — `serde_json` versus
  `std::str::parse::<f64>` on the same 40,000 literals: **1,996 differ**. Example:
  `995.6916425000001` → `995.6916425` under `serde_json`, `995.6916425000001` under both `std`
  and Python. This is pre-existing, unrelated to this round, and it matters here only because
  the Rust recorder writes `elapsed().as_secs_f64()` unrounded (`terminal/cast.rs:218`) where
  the Python recorder rounds to six decimals (`cast.py:50`), so full-precision timestamps are
  reachable. A base one ULP apart only changes a written timestamp when the scaled value sits
  within ~1e-7 of a rounding boundary, so the compound probability is negligible — but the fix
  is one word: **`serde_json`'s `float_roundtrip` feature.** I enabled it in a scratch build and
  re-ran the 40,000-literal comparison: **0 of 40,000 differ.** Worth its own change, and it
  would firm up more than this function.
- **The two rejection messages read differently for the same rejection** — Rust
  "must be at least 0.000001 seconds, got 0.0000005", Python "must be at least 1e-06 seconds,
  got 5e-07". Float formatting, not logic; pre-existing in shape. Not worth a change on its own.
- The one-directional differential harness noted in the first review is unchanged, and this
  round leans on it a little harder. Still a separate change.

## What I could not verify

- I did not test against the browser-based player, only against the engine it embeds and
  against the reference-compatible Python replay path.
- The 5% float-parse figure is measured on random doubles in a plausible session-length range,
  not on a corpus of real recordings.

md-mt:
All four addressed in `8894163`, on top of a rebase onto `e4e3de2`. Every guard below was checked by mutation rather than by reading, since that is how the last round's claims got through.

## Rebase

`CHANGELOG.md` was the only conflict, resolved by keeping both blocks as you verified. Two deviations from the literal resolution, both in ordering/prose rather than content:

- Your resolution left `### Rust — Changed` ("Nothing. …") above my `### Rust — Added`. I put Added first, per Keep a Changelog order and the rest of the file.
- That section ended "it would be a Rust behaviour change in a release whose Rust half is otherwise untouched." Merging this branch in makes that false. I reworded the clause to "it would change what Rust renders, in a release that changes no other Rust rendering behaviour" — same point, still true. Flagging it because editing someone else's entry during a conflict resolution is not something to do silently.

Post-rebase: 825 Python tests pass, corpus reproduces byte-for-byte, `mypy` clean over 42 files, version-drift clean.

## Finding 1 — the rounding was pinned by nothing

Confirmed exactly as reported. `_round_micros` → `return round(at, 6)` left the suite green **and** the corpus byte-identical, and three comments said otherwise.

I took your alternative rather than the `1.5e-6` hold, because there is a case that discriminates without an absurd input: keep the readable `0.2` hold and move the *base* to a seventh decimal, which is what a Rust-recorded cast actually ends on (`CastRecorder` writes `as_secs_f64()` unrounded).

```
base 0.5000005, hold 0.2
  transcribed : 0.700001  0.900001  1.100001  1.300001
  round(at,6) : 0.700001  0.9       1.100001  1.3
```

Three of the eight frames separate in the conformance cast. Used in both unit tests and in `session-with-fractional-hold.cast`, so it is pinned twice independently *and* across implementations.

One thing I checked before choosing that literal, given your `float_roundtrip` finding: `0.5000005` is safe from it by construction. Seven significant digits means `serde_json`'s non-roundtrip path computes `5000005.0 / 1e7` — both operands exactly representable, one correctly-rounded division — so it lands on the same double Python's parser does. The hazard needs the long literals your example had.

## Finding 2 — the floor, and the value that could not pin it

Confirmed both halves. `5e-7` is out; `6e-7` and `7.5e-7` are in, since they round up to a whole microsecond without advancing by one:

```
6e-7   accepted by round(hold,6) > 0  ->  1e-06, 1e-06, 2e-06, 2e-06
7.5e-7 accepted by round(hold,6) > 0  ->  1e-06, 2e-06, 2e-06, 3e-06
```

And `MIN_CHECKPOINT_HOLD` is now asserted as the literal `1e-6` on both sides, same as `3.0`.

Mutation results after the change, Python side:

| mutant | before | now |
|---|---|---|
| `_round_micros` → `round(at, 6)` | survived, corpus identical | **killed**, corpus differs |
| floor check → `round(hold, 6) <= 0` | survived | **killed** (3 failures) |
| `MIN_CHECKPOINT_HOLD` → `1e-3` | survived | **killed** |
| `DEFAULT_CHECKPOINT_HOLD` → `5.0` | killed | killed |
| drop `\x1b[r` | killed | killed, corpus differs |

## Findings 3–5 — the comments

- **3.** You are right and I was wrong: `round(5e-7, 6)` is `0.0`, so `5e-7` does not pass the check I claimed it passed. The nearest double sits just below the exact half. The conclusion survives — rounding up is not advancing — but on `6e-7`, not on `5e-7`, and the test comment now says exactly that, including why anything below half a microsecond cannot make the point.
- **4.** "a gap of at least 1e-6 in the input is still a gap after rounding" is now gone. The doc says what is true: it is a floor on the input, not a proof about the output; `999.9999995` at exactly the floor still collides; it needs the floor *and* a half-microsecond base together, so no recorded session reaches it, and every hold above the floor is safe at any base.
- **5.** "the microsecond floor now puts the exponent case out of reach for any accepted hold" was in my PR comment and is wrong — the floor is itself an accepted hold and is one of the cases. Correcting it here since it was the recorded reason for closing the item. The decision stands on the other half of the reasoning only: no recorder produces a session ending near zero *and* a caller asking for a sub-100-microsecond hold.
- **6.** Agreed, leaving `[true,"o","x"]`.

I also reworded the `timestamp_text` helper doc, which justified reading text with `0.3` versus `0.30000000000000004` — those differ as parsed floats too, so it was not the reason. The reason is that the bytes are the contract.

## Outside this PR

`serde_json`'s `float_roundtrip` is a good catch and I have deliberately not folded it in — it changes parse behaviour crate-wide and deserves its own change with its own before/after, not a line in a feature PR. Worth filing.

## Verification

Python: 825 tests, `ruff` clean, `mypy` clean, stdlib gate `145 passed, 6 skipped`, version-drift clean, corpus reproduces byte-for-byte. Rust: `clippy --all-targets` clean, `fmt --check` clean, `cargo doc` still 6 pre-existing warnings.

Unchanged caveat: I still cannot execute compiled binaries here, so the Rust half runs in CI. The two harness halves' `BASE_CAST` and `FRACTIONAL_BASE_CAST` literals are script-checked to decode identically. Given you can run both, the Rust mutation equivalents of A and B are the places worth a second look.

PR #167, opened 2026-08-16
Consolidate the Rust implementation into this repository (Option C)

md-mt:
Merges `md-mt/termproof-rust` into this repository as a single Option C monorepo. Both histories survive: the Rust tree is brought in with `--allow-unrelated-histories`, and nothing is rewritten, squashed or grafted.

**This PR pushes no tags and publishes nothing.** Tag work is a separate follow-up step, planned below.

## Layout

```
.github/       one CI and repository metadata surface
.gitignore     merged
spec/          the cross-implementation contract, promoted from the Rust repo's specs/
conformance/   corpus and harness, promoted from the Rust repo's harness/
python/        former md-mt/termproof root
rust/          former md-mt/termproof-rust root
```

There is deliberately **no root `docs/`**. The two documentation trees are product-, release-channel- and history-specific, and merging them mechanically manufactures stale links. `python/docs*` and `rust/docs` stay where they are; a separate editorial pass follows.

## Commits

Each is independently reviewable, and commits 1–3 change no content — a reviewer can confirm that mechanically.

| # | Commit | What it does |
|---|--------|--------------|
| 1 | `refactor(layout): move the Python implementation under python/` | Pure move. 707 paths, all `R100`, zero content changes. |
| 2 | `refactor(layout): import the Rust implementation under rust/` | Merges `termproof-rust` with `--allow-unrelated-histories`. The move under `rust/` happens in a single-parent commit **on the Rust side** (`6042496`, 178 × `R100`, off the live Rust head `bb8dc3a`), which the merge then brings in — see *Per-file history* below for why the order matters. Two parents; pure move. |
| 3 | `refactor(layout): promote the spec and conformance corpus to the root` | Pure move. `rust/specs` → `spec/`, `rust/harness` → `conformance/`. 11 paths, all `R100`. |
| 4 | `ci: consolidate both CI surfaces at the repository root` | The substantive change. Details below. |
| 5 | `chore: merge the remaining root-level repository configuration` | Root `.gitignore`, Dependabot ecosystems, CODEOWNERS. |

Rename-only proof for 1–3:

```
git show --name-status -M <sha> | grep -v '^R'   # empty for each
```

### What commit 4 does

GitHub reads only the root `.github`, so the two implementations cannot keep their own.

- Workflows namespaced `python-*` and `rust-*`, each with a **path filter**, so a change to one implementation does not run the other's CI.
[...]

PR #195, opened 2026-08-20
fix(svg)!: one source of truth for the SVG geometry defaults, on a Linux-safe font stack

md-mt:
## What this is

`SvgStyle` (`python/termproof/attributed.py`) and `SvgRenderConfig`
(`python/termproof/config.py`) both describe SVG geometry, and their defaults
disagreed in **every single field**. A consumer moving from
`screen_svg(screen, SvgStyle())` to a configured `SvgRenderer` — the natural
move, and the one the API shape invites — silently resized, re-padded and
re-coloured every image it had ever produced. Nothing failed.

`SvgStyle` is now the single canonical definition. `SvgRenderConfig` takes each
default from the same `DEFAULT_*` constant instead of restating it, and a
field-by-field test asserts that an unconfigured
`SvgRenderConfig().style(cols, rows)` *is* `SvgStyle(columns=cols, rows=rows)`.

**This changes rendered output.** A 120x40 screen was 1116x836 and is now
1220x900. `CHANGELOG.md` names every default that moved and gives the
pin-to-keep-old-output YAML.

| field | was (`SvgRenderConfig`) | is (canonical, from `attributed.py`) |
|---|---|---|
| `char_width` / `cell_w` | `9` | `10.0` — `DEFAULT_CELL_W` |
| `line_height` / `cell_h` | `20` | `22.0` — `DEFAULT_CELL_H` |
| `font_size` / `font_px` | `14` | `16` — `DEFAULT_FONT_PX` |
| `padding` | `18` | `10` — `DEFAULT_PADDING` |
| `bg` | `#101418` | `#0b0f14` — `DEFAULT_BG` |
| `font_family` | `ui-monospace,SFMono-Regular,Menlo,Consolas,monospace` | `Noto Sans Mono, Liberation Mono, monospace` — `FONT_STACK` |
| `fg` | `#e6edf3` | `#e6edf3` — unchanged |
| floors | `min_width: 320, min_height: 160` | none, on both paths |

## The three judgement calls

**Font stack — `SvgStyle`'s wins, as the brief directed.** No argument to make:
the old stack named three fonts that exist on macOS and Windows and none that
exist on a stock Linux image, which is where CI renders. It collapsed to generic
`monospace`, and `Menlo` resolves to *proportional* DejaVu Sans on some images.
`screen_svg` places one `<text>` per cell at `x = column * cell_w`, so a
substituted glyph cannot shift its neighbours — but a proportional face still
overflows its cell and overlaps the next one. Two tests pin the ordering and pin
that the four off-Linux names are absent.

**Geometry and colour — `SvgStyle`'s are canonical.** I looked for evidence that
`SvgRenderConfig`'s were deliberate and found none still true. The comment above
them says they "reproduce the behavior that was previously hardcoded in the
renderers", i.e. they were chosen for byte-compatibility with a pre-config
renderer, not for quality; `git log -S` puts all of them in the squashed
consolidation commit (#167) with no independent rationale; and neither
`python/README.md` nor `docs/evidence-quality.md` states a contract on the
values, they only restate them as examples (both updated here).

One incidental win: with `cell_w: 10, padding: 10`, canvas width is
`cols * 10 + 20` — always even. The old `cols * 9 + 36` was odd for odd `cols`,
which is exactly what `scale=trunc(iw/2)*2` exists to paper over in the video
pipeline.

**Floors — both paths are floorless.** `style()` set `min_width: 320,
min_height: 160`; `SvgStyle`'s own defaults were `0`. Reasons for choosing zero:

1. Rust's `SvgMetrics` has no floor concept at all, so zero is the only choice
   that makes the two implementations agree.
2. An SVG is resolution independent. A floor does not add information to a small
   grid, it surrounds it with dead background, and it costs the otherwise exact
   `canvas == grid + 2 * padding` relation that every consumer of
   `width`/`height` assumes.
3. `320x160` is copied from `PngRenderer`, where it *is* justified and stays: a
   PNG is a fixed pixel count, so a 40x60 one really is unreadable. That reason
   is true of raster and not of vector.

`min_width`/`min_height` stay on `SvgStyle` (public API; removing them would be
a breaking change for no gain) for a caller that rasterises at intrinsic size
and wants a floor. A test covers that path still working.

## Rust: the finding you asked for

**The same double-default does not exist in Rust, and no Rust change was needed
or made** (`git diff origin/main -- rust/` is empty).

- `SvgMetrics::default()` in `terminal/attributed.rs` already reads every value
  from the `DEFAULT_*` constants declared beside it.
- Its two other constructors, `ScreenshotRenderer` and `CastVideoConverter`,
  already `use crate::terminal::attributed::DEFAULT_*` rather than restating
  numbers, and `evidence/render.rs::metrics` builds from
  `..Default::default()`.
- `FONT_STACK` there is already the Linux-first one, with a comment saying why:
  *"the boxes render on a CI host that has neither Menlo nor Consolas, so naming
  them first just wastes a lookup."*
- Rust has no `min_width`/`min_height`, which is the behaviour both sides now
  have.

So Python moved *onto* the values Rust has been rendering all along; the two
implementations agree on SVG geometry for the first time.

I did not extend the conformance pair. `probe_evidence_manifest.py` and
`differential_evidence_manifest.rs` stub the renderer deliberately (the real one
shells out to `rsvg-convert`, whose version is a property of the machine), so
SVG geometry is not a surface it compares, and this change adds no new surface —
it removes a divergence. Cross-language pinning is instead the two languages'
constants being literally the same numbers, with
`tests/test_svg_geometry_defaults.py::CanonicalConstantsTest` pinning them on the
Python side. I ran the manifest probe: the recorded corpus is unchanged.

## The corpus regeneration

156 SVG files under `python/examples/artifacts/` are regenerated, one line each.
`CorpusByteIdentityTest`'s docstring requires exactly this of any change to the
defaults — *"has to be accompanied by a regenerated corpus, which puts the new
bytes in the diff where a reviewer can see them"* — and that is the point of the
gate. They were re-rendered through the same path the gate replays (`final.svg`
from the attributed cast replay, step images from their recorded text), so the
gate reproduces them byte for byte.

## Two things slightly outside the brief, easy to back out

1. **`evidence.png.fg`/`bg` also move onto `DEFAULT_FG`/`DEFAULT_BG`** (bg
   `#101418` → `#0b0f14`). Not doing this would have manufactured a *new*
   disagreement: the PNG and SVG screenshots of one run would have rendered
   different terminal backgrounds, where today they match. `png.padding`,
   `png.font_size` and `png.scale` are deliberately untouched — they are raster
   quantities with no SVG counterpart, since that renderer measures its cell off
   the PIL face rather than off a cell grid.
2. **`BUILTIN_DEFAULTS["evidence"]` is derived with `asdict()`** rather than
   restating each default a *third* time. `python/README.md` points at that dict
   as the documentation of every knob, so a value there that disagrees with the
   one a renderer applies is a promise the code does not keep. This is the first
   commit and is behaviour-preserving on its own (the produced dict is identical
   key for key).

Not touched: `visual_diff.py` / `evidence/diff.rs` still use the macOS stack for
the `baseline` / `actual` captions on the comparison page. Same Linux problem,
but it is chrome text on a light background, not a column grid, so a
proportional fallback there only changes a label's width. Worth a separate PR;
out of scope for this one.

## Commits

Split per Tidy First — structural first, behavioural second:

1. `refactor(config)` — move the config dataclasses above `BUILTIN_DEFAULTS` and
   derive the evidence block from them. No behaviour change; the produced dict
   is identical and the suite is untouched.
2. `fix(svg)!` — converge the defaults, drop the floors, regenerate the corpus.

## Test plan

- `uv run python -m unittest discover -s tests` — **791 passed** (779 before;
  +12 new).
- New `python/tests/test_svg_geometry_defaults.py`: pins each canonical constant
  by literal value; pins each `SvgStyle` and each `SvgRenderConfig` default
  field by field against the constant; asserts `SvgRenderConfig().style()`
  equals `SvgStyle()` at three grid sizes including 3x2 where floors used to
  bind; asserts every `SvgStyle` field except the deliberate four is reachable
  from `SvgRenderConfig`; asserts the font stack ordering and the absence of the
  off-Linux names; asserts PNG shares the palette.
- Mutation-checked: reverting `padding` to `18` in `SvgRenderConfig` fails 6 of
  the 12 new tests.
- `uv run ruff check` clean, `uv run mypy termproof` clean.
- `cargo test --manifest-path rust/Cargo.toml --workspace` — 25 test binaries,
  all pass, no failures.
- Conformance: `probe_evidence_manifest.py` output still byte-identical to
  `conformance/corpus/evidence_manifest.expected.json`.

No version bump — whether this ships as a patch or a minor is the captain's
call. The entry is under `[Unreleased]`, split `Python — Changed` /
`Rust — Changed` per the file's convention.


md-mt:
Review addressed — both findings you flagged are fixed in `e9ca247`, plus one you did not name and finding 4.

## Finding 1 (major) — fixed, and it was worse than reported

You were right, and the reasoning was exactly right: the floor belonged to the raster path, not the vector one, and I removed it from the wrong side. `RsvgPngRenderer` calls `rsvg-convert --output <png> <svg>` with no `-w`/`-h`/`-z`, so intrinsic size *is* pixel size.

**There is a third consumer with the same defect that the review did not name.** `RsvgFfmpegBackend.render` in `cast_video.py:248` built its frame geometry from the same `svg_config.style(columns, rows)` and rasterises every frame through `rsvg-convert` the same way — so a small cast became a small video, not just a small screenshot. Fixed alongside.

The fix is `SvgRenderConfig.raster_style()`: `style()` floored at `RASTER_MIN_WIDTH` x `RASTER_MIN_HEIGHT`, applied as a hard lower bound exactly the way `PngRenderer` has always applied those two numbers. Both rsvg consumers use it. And since a restated constant is the bug this PR exists to remove, `PngRenderer`'s own `max(320 * scale, …)` now reads the same two names — one definition, three raster renderers.

Reproducing your table at the new head:

```
png_rsvg 120x40 -> 1220x900
png_rsvg  20x4  -> 320x160     (was 220x108 at 39398b1; 320x160 on main)
png_rsvg   3x2  -> 320x160     (was 50x64)
```

I took your suggestion on the test too. `test_an_explicit_floor_still_works_for_a_caller_that_wants_one` was testing an unreached branch, as you said. There is now a `RasterFloorTest` with six cases, and the two that matter drive the *real* renderers with a fake rasteriser and read the markup on its way out, rather than asserting on a style object. Reverting either `raster_style` call fails them:

```
AssertionError: 'width="320"' not found in '<svg … width="30" height="42" …>'
Ran 18 tests … FAILED (failures=2)
```

## Finding 2 (minor) — fixed by making the recipe actually exact

Rather than only softening the wording, I added `min_width`/`min_height` as real `evidence.svg` keys defaulting to `0` — the resolution you suggested. Three things fall out:

- The recipe is now exact at every grid size, verified by rendering:

  | grid | main, unpinned | PR head, recipe applied |
  |---|---|---|
  | 120x40 | 1116x836 | 1116x836 |
  | 20x4 | 320x160 | 320x160 |
  | 3x2 | 320x160 | 320x160 |

- The CHANGELOG says when the two extra keys are needed (below 32 columns or 7 rows) instead of claiming the six alone are enough.
- It closes the last `SvgStyle` ↔ `SvgRenderConfig` gap. `test_every_style_field_is_reachable_from_the_render_config` no longer exempts `min_width`/`min_height`; only `columns`/`rows` remain, and those are the grid being rendered rather than a configured value.

## Finding 4 (minor) — fixed

`must be at least 1, got 0.5`. You are right that softening the wording and widening the type in the same commit interacted badly, and right that eight int fields lost the word "integer" they were entitled to — naming the bound fixes both at once, since the type is `_check_field_type`'s job and it reports separately.

## Finding 3 (minor) — not changing the stack; naming the cost

I am taking the second of your two options, explicitly, because you are right that doing it silently is the objectionable part. The reason is the one you flagged as coupling: Rust's `FONT_STACK` is the same literal string, and this PR's strongest evidence is that Python and Rust now emit byte-identical SVG for the same input. Appending the four off-Linux names is a coupled two-language change, and it is a good one — just not this one.

The cost is now written down in `test_the_stack_names_nothing_that_only_exists_off_linux`'s docstring rather than left implied: **off-Linux viewers get generic `monospace` where they previously got SF Mono, Menlo or Consolas.** That is a real downgrade for most humans reading the published evidence, and the PR should not have presented the change as pure gain.

## One correction to my own CHANGELOG

Writing up the raster fix I claimed Rust's `ScreenshotRenderer` passes `-w`/`-h` explicitly. It does not — `screenshot.rs:178` invokes `--output <png> <svg>`, same as Python, and `CastVideoConverter` likewise. So **Rust's raster path has never had a floor**, and Python's has since it existed: the two already disagreed here before this PR. Restoring Python's floor restores that status quo rather than widening the gap. The CHANGELOG now says exactly that, and flags giving `SvgMetrics` a matching raster floor as worth doing and deliberately out of a release whose Rust half is otherwise empty. `git diff origin/main -- rust/` is still empty.

## Verification at `e9ca247`

- `unittest discover -s tests` → **797 passed** (791 at your review, +6 raster).
- `ruff` clean, `mypy termproof` clean, `check_version.py` and `check_schema_copies.py` pass.
- Corpus byte-identity gate unaffected: `SvgRenderer` still uses `style()` and the floor never binds at 120x40, so the 156 regenerated files are unchanged by this commit.
- PNG bytes unchanged by the constant extraction — `RASTER_MIN_WIDTH * scale` is `320 * scale`.

## Your two process notes

Both acknowledged and neither is mine to resolve:

- **Version policy** — still `0.4.0`, still the captain's call, and your reading that the stated `0.x` rule points at `0.5.0` is on the record here. I could not register the hold either; the same sandbox deny that blocked your `firstmate/state` writes blocks mine, so `tp-fb-svg-a2.status` has never been written for this task and my reports have gone in chat instead.
- **`python/site/artifacts/` being two generations stale** and **the missing CI baseline report** are both real, both pre-existing, and both worth issues. I have not touched either — the first has no gate forcing regeneration, which as you say is arguably the actual bug.

md-mt:
# PR 195 — fix(svg)!: one source of truth for the SVG geometry defaults, on a Linux-safe font stack

**VERDICT:** CHANGES NEEDED
**MERGE RISK:** medium — the `png_rsvg` renderer loses its 320x160 canvas floor on a
*raster* path, which is the exact case the PR's own text says a floor is for; below
~29 columns or ~6 rows it now emits a postage-stamp PNG where it used to emit a
readable one.
**CI:** 12 passed, 0 failed, 2 skipped, 14 total — green, and ran 13 minutes before
this review against the current base. The two skips are the Rust workflows, correctly
path-filtered out because `rust/` is untouched (verified: `git diff f033dcc pr195 -- rust/`
is empty). Coverage for the *code* change is strong — the full 791-test suite runs,
including the 156-file byte-identity corpus gate. Coverage for the *claim* is weaker
in two specific ways, both noted below: no CI job rasterises an SVG through fontconfig,
so the "Linux-safe font stack" rationale is not exercised anywhere in Python CI; and
the repo's own before/after evidence comparison produced no base report, so CI did not
show any reviewer the pixel delta (pre-existing — see "Outside this PR").
**STALENESS:** Not stale. `git merge-base --is-ancestor main pr195` succeeds and the
merge base *is* `main`'s head `f033dcc`, so this is a fast-forward. Full suite, ruff,
mypy, `check_version.py` and `check_schema_copies.py` all pass locally at PR head.

## Summary

The central claim is true and I verified it by rendering rather than by reading. Two
types did describe the same SVG geometry with defaults that disagreed in every field,
`SvgStyle` is now genuinely the single definition, and `SvgRenderConfig` genuinely
references it rather than restating it. The convergence is real enough that the Python
configured renderer and the Rust renderer now emit byte-identical SVG for the same
input, which they did not before. The regression test is one of the better ones I have
reviewed: every single-field mutation I could construct is caught.

Two things are wrong. One is a behaviour defect the PR introduces on a path its own
reasoning exempts (`png_rsvg`), and one is a migration recipe that the CHANGELOG says
reproduces the old output "exactly" and which I ran and found does not, below the old
floor. Both are small fixes. Neither undermines the change's direction.

---

## Findings

### 1. `png_rsvg` is a raster renderer and it just lost its canvas floor — major

- **severity:** major
- **location:** `python/termproof/config.py:70-90` (`SvgRenderConfig.style`, floors
  removed) reaching `python/termproof/rsvg.py:107` and `:113-118`

`RsvgPngRenderer` builds its markup from `self.config.style(cols, rows)` — the same
`SvgRenderConfig` this PR made floorless — and then invokes
`rsvg-convert --output <png> <svg>` with **no `-w`, `-h` or `-z`**. rsvg-convert
rasterises at intrinsic size, so the PNG's pixel dimensions are exactly the SVG's
`width`/`height` attributes. Removing the floor from `style()` therefore removes it
from a PNG.

The PR argues the floor should go because "an SVG is resolution independent — a viewer
scales it," and argues in the same breath that `PngRenderer` keeps its floor because
"a PNG is a fixed pixel count, so a 40x60 one really is unreadable." Both sentences are
correct. The problem is there are two PNG renderers and only one of them kept the floor.
`python/termproof/config.py:82-89` and `CHANGELOG.md:85-94` both now assert the raster
path stays floored; for `png_rsvg` that is a promise the code does not keep — which is
precisely the failure mode this PR was written to eliminate.

The escape hatch the PR deliberately preserved does not reach here either. The PR keeps
`min_width`/`min_height` on `SvgStyle` "for a caller that rasterises at intrinsic size
and wants a floor" (`python/termproof/attributed.py:329-333`). `RsvgPngRenderer` is
exactly that caller, it is in-tree, and it takes a `SvgRenderConfig` — which has no key
for either field. The one consumer the exception was written for cannot use it.

**Verified, by running it.** I replaced the `ToolRunner` with a capture and read the
`width`/`height` off the markup that reaches `rsvg-convert`:

```
=== PR HEAD ===
png_rsvg 120x40 grid -> raster 1220x900 px
png_rsvg  20x4  grid -> raster  220x108 px
png_rsvg   3x2  grid -> raster   50x64  px
=== MAIN (f033dcc) ===
png_rsvg 120x40 grid -> raster 1116x836 px
png_rsvg  20x4  grid -> raster  320x160 px
png_rsvg   3x2  grid -> raster  320x160 px
```

**Concrete failure case:** a run with a 20x4 grid and `screen_renderers: png_rsvg`
produced a 320x160 PNG before this change and produces a 220x108 PNG after it. Nothing
errors; the evidence image is just smaller and harder to read, and no test notices —
`test_rsvg.py` and `test_cast_video.py` contain no reference to `320`, `160`,
`min_width` or `min_height`.

**What I would do:** floor the raster path, not the vector one. Either give
`SvgRenderConfig` `min_width`/`min_height` keys defaulting to `0` and have
`RsvgPngRenderer.from_config` apply `320`/`160` (mirroring `PngRenderer`), or apply the
floor inside `RsvgPngRenderer.render_attributed` via `dataclasses.replace` on the style
it just built. Then extend `test_svg_geometry_defaults.py::CanvasGeometryTest` with a
case pinning that `png_rsvg` at a small grid is still floored — the current
`test_an_explicit_floor_still_works_for_a_caller_that_wants_one` pins the mechanism but
no production caller uses it, so it is currently testing an unreached branch.

### 2. The CHANGELOG's pin-to-keep-old-output recipe does not, below the old floor — minor

- **severity:** minor
- **location:** `CHANGELOG.md:62-74`

The recipe is introduced with "To keep the old output **exactly**, pin all six in
`.termproof/config.yaml`". I followed it literally, byte for byte, and rendered the same
input on `f033dcc` and on PR head through `load_config`.

**Verified, by running it.** At 120x40 the recipe works perfectly — identical sha,
identical `1116x836` canvas. At 3x2 it does not:

| grid | main, unpinned | PR head, recipe applied | match |
|---|---|---|---|
| 120x40 | 1116x836, sha `…` | 1116x836, same sha | yes |
| 3x2 | 320x160 | **63x76** | no |

The six pinned keys restore the cell metrics and the palette; they cannot restore the
floors, because `SvgRenderConfig` has no YAML key for `min_width`/`min_height` and this
PR did not add one. The floor removal *is* documented, thoroughly, in the very next
CHANGELOG bullet (`CHANGELOG.md:85-94`) — but that bullet never says the recipe above
it is incomplete, and a reader following the recipe has been told it is exact.

**Concrete failure case:** the floor bound whenever `cols*9 + 36 < 320` (cols ≤ 31) or
`rows*20 + 36 < 160` (rows ≤ 6). A consumer on a small grid who applies the documented
recipe still gets differently-sized artifacts and has been told they would not.

**What I would do:** one sentence. Change "To keep the old output exactly" to "To keep
the old cell metrics and palette exactly" and add "the 320x160 floor cannot be pinned
back; see the next entry." If finding 1 is fixed by adding `min_width`/`min_height` YAML
keys, the recipe becomes genuinely exact and this resolves itself.

### 3. The font-absence test forbids a stack that would be strictly better — minor

- **severity:** minor
- **location:** `python/tests/test_svg_geometry_defaults.py:165-170`

`test_the_stack_names_nothing_that_only_exists_off_linux` asserts `ui-monospace`,
`SFMono-Regular`, `Menlo` and `Consolas` are **absent** from the stack. The stated
rationale only requires that they not come *first* — and the sibling test
`test_the_canonical_stack_names_linux_fonts_before_the_generic_fallback` already pins
that ordering correctly. Absence is a stronger constraint than the rationale supports,
and it locks out `"Noto Sans Mono, Liberation Mono, ui-monospace, SFMono-Regular, Menlo,
Consolas, monospace"`, which would satisfy every argument the PR makes on Linux *and*
give a macOS or Windows reader a real terminal face instead of the browser's generic
`monospace` default.

That matters more than it sounds, because the published evidence SVGs are read in
browsers. **Inferred, not measured:** on any macOS or Windows viewer neither
`Noto Sans Mono` nor `Liberation Mono` is present, so the published screenshots now
resolve to the generic `monospace` alias where they previously resolved to SF Mono,
Menlo or Consolas. That is a small visual downgrade for the majority of humans looking
at the evidence, and the PR does not name it as a trade-off — it presents the change as
pure gain.

**What I would do:** relax the assertion to "none of these four appears before
`Noto Sans Mono`", or keep the absence test and say in the PR body that off-Linux
viewers are deliberately being given generic monospace for cross-language string parity
with Rust. Either is defensible; silently doing the second while arguing the first is
what I would push back on. Note this is coupled: Rust's `FONT_STACK` is the same literal
string, so any change here needs both sides.

### 4. `must be positive, got 0.5` is false — minor

- **severity:** minor
- **location:** `python/termproof/config.py:349-360` and `:335-336`

Widening `char_width`/`line_height` to `float` made fractional values reach the range
check for the first time, and the same commit softened its wording from "must be a
positive integer" to "must be positive". The minimum is still `1`, so the two changes
interact badly.

**Verified, by running it:**

```
PR head: REJECTED: evidence.svg.char_width must be positive, got 0.5
PR head: REJECTED: evidence.svg.line_height must be positive, got 0.9
main:    REJECTED: evidence.svg.char_width must be int, got 0.5
```

`0.5` is positive. On `main` the message was correct because the type check caught it
first. Now the user is told a rule that their value satisfies, and is never told the
real one. `char_width: 9.5` loads fine, so this is a reachable configuration space, not
a hypothetical.

No test pins the wording — `test_config.py:415-436` asserts only that the key name
appears — so this is invisible to the suite.

**What I would do:** make the message state the bound: `must be at least 1, got 0.5`.
Note the wording was softened for the two float fields but applies to eight int fields
(`png.scale`, `png.font_size`, `png.padding`, `video.fps`, `video.fps_cap`,
`video.font_size`, `svg.font_size`, `svg.padding`), all of which lost the word "integer"
they were entitled to.

---

## Evidence: what I actually ran

**Full suite at PR head** — `uv run python -m unittest discover -s tests` →
`Ran 791 tests in 101.407s / OK`. Matches the PR's claim of 791 (779 before, +12).
`uv run ruff check` → "All checks passed!". `uv run mypy termproof` → "Success: no issues
found in 41 source files". `scripts/check_version.py` and `scripts/check_schema_copies.py`
both pass.

**Blast radius, rendered before and after** (`/tmp/render_probe.py`, run on `f033dcc`
then on `39398b1`, output diffed):

| entry point | before | after | changed? |
|---|---|---|---|
| `screen_svg(screen, SvgStyle())` | sha `63183d62702dae2e` | sha `63183d62702dae2e` | **no — byte-identical** |
| `SvgRenderer()` / `screen.render_svg` @120x40 | 1116x836, sha `c71e242b034c02d3` | 1220x900, sha `63183d62702dae2e` | **yes** |
| `SvgRenderer()` @3x2 | 320x160 | 50x64 | **yes** |
| `RsvgPngRenderer` @20x4 | 320x160 px | 220x108 px | **yes** (finding 1) |
| `PngRenderer` | `#101418` page | `#0b0f14` page; raster geometry and 320x160 floor unchanged | **colour only** |
| `AggFfmpegBackend` video | corpus mp4s measure 1078x962 / 982x290 | unchanged | **no** — agg has its own metrics, unrelated to `SvgMetrics` |
| Rust `screen_svg` | unchanged | unchanged | **no** — `rust/` diff is empty |
| a caller pinning all six in YAML | — | identical at grids above the old floor | **no**, above the floor |

The middle row is the whole point of the PR, and it lands: after the change the
configured renderer's sha `63183d62702dae2e` is *the same sha* the direct `SvgStyle()`
path produces. That is the convergence, demonstrated rather than asserted.

No in-repo recipe or config pins `evidence.svg` — the only checked-in
`.termproof/config.yaml` (`python/plugin-template/examples/`) sets plugin references
only — so every in-tree consumer takes the new geometry, which is what the corpus
regeneration reflects.

**Is the single source of truth real?** Yes. `SvgRenderConfig`'s seven fields each
default to a `DEFAULT_*` constant imported from `attributed.py`
(`python/termproof/config.py:61-67`); `BUILTIN_DEFAULTS["evidence"]` is derived with
`asdict()` (`:227-232`) rather than restated a third time; `python/README.md` and
`python/docs/evidence-quality.md` were both updated. I grepped for every restatement of
the old values and of the geometry keys: the only surviving literals are `PngRenderer`'s
own raster constants (`builtin_renderers.py:89-93`, deliberate and documented), a
`before-monochrome.svg` fixture, `visual_diff.py:154` (the PR explicitly scopes this
out — accurate), and `python/site/artifacts/` (pre-existing; see below). This is a real
convergence, not two copies with equal values.

**Would the regression test fail on a single-field change?** Yes — every time. I
mutated one default at a time and re-ran both gates:

| mutation | `test_svg_geometry_defaults` | corpus byte gate |
|---|---|---|
| `SvgRenderConfig.padding` → 18 | FAILED (6 of 12) | FAILED |
| `char_width` → 9.0 | FAILED (6) | FAILED |
| `font_size` → 14 | FAILED (4) | FAILED |
| `line_height` → 20.0 | FAILED (6) | FAILED |
| `bg` → `#101418` | FAILED (5) | FAILED |
| `font_family` → macOS stack | FAILED (5) | FAILED |
| `style()` floors restored to 320/160 | FAILED (4) | **OK** |
| `DEFAULT_PADDING` constant → 12 | FAILED (3) | FAILED |

The PR claims "reverting `padding` to `18` fails 6 of the 12 new tests" — exactly right.
The seventh row is the interesting one: the corpus gate is *blind* to the floors, because
they never bind at 120x40. That is the specific hole the new test exists to close, and
it closes it. Every field that disagreed is covered, including the font stack and the
floors.

**The font stack, checked against the images the project actually renders on:**

- **GitHub `ubuntu-latest`, Python CI** — I grepped every workflow: nothing installs
  `librsvg2-bin`, `rsvg-convert`, or any font package. The verify job writes SVG *text*
  and renders video through `agg`, which carries its own font. **No Python CI job
  resolves `FONT_STACK` through fontconfig at all.** The "CI renders on Linux where the
  previous stack collapsed" framing is therefore not about Python CI. It is not wrong —
  it is about the Rust image and about downstream rasterisers — but it is doing less
  work than the PR body implies.
- **`rust/docker/termproof.Dockerfile`** — installs `fonts-liberation2` and
  `librsvg2-bin`, with a comment naming this exact reasoning (lines 51-57). So the
  second entry, `Liberation Mono`, resolves; the **first entry, `Noto Sans Mono`, is not
  installed in any image in this repository.** The ordering does what is claimed (falls
  through to a real monospace face), but the stack is effectively one-deep on the only
  image that rasterises, and if `fonts-liberation2` were ever dropped the generic
  `monospace` alias would have nothing to resolve to. Worth knowing; not a defect here.
- **`python/docker/termproof.Dockerfile`** — installs no fonts and no rasteriser, so it
  cannot resolve the stack and cannot run `png_rsvg` (which raises a clear
  `RuntimeError`). Not affected.
- **Can the *new* stack resolve to a proportional face?** On any image in this repo, no:
  either `Liberation Mono` matches exactly, or the last entry is the generic `monospace`
  alias, which fontconfig maps to a monospace face by definition. The new stack is
  strictly safer than the old one on this axis, because the old one led with four family
  names fontconfig has no alias for.
- **What I could not verify:** whether `Menlo` in fact resolves to *proportional* DejaVu
  Sans under `rsvg-convert` on a stock image. There is no `rsvg-convert` on this machine
  and the brief forbids Docker. The mechanism is consistent with how `FcFontMatch`
  behaves on an unknown family (it never fails; it returns a best match), but I am
  labelling it **inferred, not measured**. It does not change the conclusion: the new
  stack is safe regardless of whether the old one was unsafe for that specific reason.

**The floors, rendered small enough to bind:** confirmed above — 3x2 goes 320x160 →
50x64, exactly as the CHANGELOG says, and `SvgRenderConfig().style(3,2) ==
SvgStyle(columns=3, rows=2)` now holds. The floors did win in favour of zero, and the
result matches the documentation on the SVG path. The raster path is finding 1.

**Rust:** `git diff f033dcc pr195 -- rust/ conformance/ spec/` is empty. Every Rust
constant is literally equal to Python's canonical value (`DEFAULT_CELL_W 10.0`,
`DEFAULT_CELL_H 22.0`, `DEFAULT_FONT_PX 16`, `DEFAULT_PADDING 10`, `DEFAULT_FG #e6edf3`,
`DEFAULT_BG #0b0f14`, and the identical `FONT_STACK` string). `SvgMetrics` has no
`min_width`/`min_height` anywhere in `rust/crates/`, and `ScreenshotRenderer`,
`CastVideoConverter` and `evidence/render.rs::metrics` all build from those constants or
from `..Default::default()`. The PR's Rust section is accurate in every particular.

I went one step further than the PR did and rendered the same 3x2 input through both
implementations. Built a throwaway integration test against `termproof::terminal::
attributed::screen_svg` and byte-compared its output with Python's:

```
rust == python AFTER : True    (426 bytes each, identical)
rust == python BEFORE: False
```

So this PR does not leave the two implementations further apart — it makes the Python
*configured* renderer byte-identical to Rust for the first time. The scratch test was
deleted; the tree is clean.

**Conformance:** I re-ran `probe_evidence_manifest.py` and its output is byte-identical
to `conformance/corpus/evidence_manifest.expected.json`, as claimed. I also checked *why*
that is weak evidence rather than strong: the expected corpus contains zero occurrences
of the string `svg`, so SVG geometry genuinely is not a surface it compares. The PR says
this itself and is right to.

---

## What is good

**The corpus regeneration is the right call and its diff is genuinely informative.** One
line changed in each of 156 files because each SVG is a single line, and reading one of
them shows something the PR body undersells: on `main`, the palette's `black`
(`#0b0f14`) differed from the page background (`#101418`), so every black cell emitted a
redundant `<rect>` and the rendered terminal had a subtly two-tone background. After the
change those cells match the page and are elided. `colour-stress/final.svg` goes from
opening with four black rects to opening with the first *real* colour. The palette
convergence removed a visual artifact, not just a number mismatch.

**The mutation table above is the author's claim, verified.** Twelve new tests, every
single-field revert caught, and the one mutation the byte-identity gate misses is
exactly the one the new test was written for. That is a test suite that knows what it is
for.

**The Tidy First split is real.** Commit `562bfcf` moves the dataclasses above
`BUILTIN_DEFAULTS` and derives the dict with `asdict()` — structural, behaviour-
preserving, produced dict identical key for key. `39398b1` is the behaviour change.
Reviewing them separately actually works.

**The `cols*10 + 20` evenness observation is correct.** I checked the arithmetic: the new
canvas width is always even and the new height is always even, where `cols*9 + 36` was
odd for odd `cols`. That genuinely removes a class of problem the video pipeline's
`scale=trunc(iw/2)*2` exists to paper over. Small, but the author noticed it and it holds.

**The judgement calls are argued from evidence rather than taste**, and the argument
survives checking. I looked for the thing the PR says it looked for — a reason
`SvgRenderConfig`'s values were deliberate — and found the same nothing: the comment
above them claimed byte-compatibility with a pre-config renderer, and no doc states a
contract on the values. The three documented rationales for choosing zero floors are
sound on the vector path. Only the raster path was missed.

**The PR is honest about its own scope.** `visual_diff.py` still using the macOS stack
for caption chrome is called out and correctly deferred; the conformance harness's
blindness to SVG geometry is stated rather than glossed; the two out-of-brief changes are
flagged as easy to back out. All three of those statements check out.

---

## For the maintainer — version policy, not decided here

`CHANGELOG.md:8` states the project rule: "under `0.x` a breaking change bumps the minor
digit." The title carries `!`, the change alters every rendered artifact, and the version
is unchanged at `0.4.0` across `pyproject.toml`, `rust/Cargo.toml` and `CHANGELOG.md`.
The PR says explicitly: "No version bump — whether this ships as a patch or a minor is
the maintainer's call."

Two things bear on that decision, and I am reporting them rather than resolving them:

1. `scripts/check_version.py` **does not catch this** — I ran it at PR head and it
   reports "version 0.4.0 consistent across pyproject.toml, rust/Cargo.toml and
   CHANGELOG.md". The gate checks agreement between the three files, not that a breaking
   entry under `[Unreleased]` implies a minor bump. So nothing mechanical will stop this
   shipping as `0.4.1`.
2. The version train is shared between the two implementations, and `rust/` changed
   nothing. A minor bump moves Rust's number for a Python-only behaviour change. The
   CHANGELOG's own "How to read this file" section anticipates exactly this ("a version
   can carry entries for an implementation whose own artifact has not been published at
   that number yet"), so it is consistent with the documented model — but it is still a
   real consequence worth stating out loud.

My reading of the stated rule is that it points at `0.5.0`. I am not making that call.

---

## Outside this PR

Real, separate, and must not colour the verdict.

**The repo's before/after evidence comparison produces no baseline, on every open PR.**
The CI report comment on PR 195 says "Before: base commit `f033dcc…` — No TermProof
report was generated," and consequently reports all seven recipes as `SKIP -> PASS`,
which is an artifact of the missing baseline rather than a behavioural delta. I checked
PRs 193 and 194: identical, same base, same message. This is pre-existing and not this
PR's regression — but it is why I had to render the before/after by hand. The one
mechanism in this project designed to show a reviewer a pixel delta showed nothing on
the one PR in the current set that changes pixels. Worth an issue.

**`python/site/artifacts/` is the curated evidence pack published to Pages
(`pages.yml`, asserted by `test_docs_pages.py:259-264`) and it is two generations
stale.** Those files still carry the *pre-attributed* markup shape
(`<style>text{font:14px ui-monospace,…}` with one `<text>` per line) and the old
palette; `git log` puts their last touch at #167. Already wrong before this PR. This PR
does not make it wrong, but it does widen the gap — the published site will now differ
from real output in geometry and palette as well as markup shape. Regenerating it has no
gate forcing it, unlike `examples/artifacts/`, which is arguably the actual bug.

**The corpus byte-identity gate is structurally tautological** and its docstring says so
honestly: the corpus is produced by the renderer the gate then compares against, so it
can only catch an *undeclared* change, never a *wrong* one. That is a legitimate design
— putting the bytes in the diff is the point — but it means the gate passing on this PR
is not evidence the new geometry is correct, only that it was declared. The new
`test_svg_geometry_defaults.py` is what carries the correctness claim, and it does carry
it.

md-mt:
# PR 195 — re-review after the fix round (`39398b1` → `e9ca247`)

**VERDICT:** MERGE WITH NITS
**MERGE RISK:** low — the two findings that carried the previous CHANGES NEEDED are
fixed, and I verified both by rendering rather than by reading. What is left is two
documentation/test-coverage nits and a mechanical CHANGELOG rebase.
**CI:** green on the current head — 12 passed, 0 failed, 2 skipped, 14 total. The two
skips are the Rust workflows, correctly path-filtered out: `git diff f033dcc e9ca247 --
rust/ conformance/ spec/` is empty. Note the run was against merge base `f033dcc`, which
is no longer the tip of the default branch; I ran the merged tree locally instead (below).
**STALENESS:** **This no longer merges cleanly.** Two changes have landed on the default
branch since the merge base (`8393db0`, `fc8adbf`), and `CHANGELOG.md` conflicts — both
sides replace the same `Nothing yet.` line under `[Unreleased]`. `python/README.md`
auto-merges. I resolved the conflict by keeping both entry sets and ran the merged tree:
**811 tests OK**, `ruff` clean, `mypy` clean. So the conflict is a rebase chore, not a
semantic collision.

## Summary

Scope of this pass: only the delta `39398b1..e9ca247` (9 files, +279/−47), plus a
re-run of the two failure cases the previous review reproduced. Everything the previous
review cleared and the delta does not touch — the single-source-of-truth convergence,
the corpus regeneration, the Tidy First split, the Rust half being untouched, the
cross-implementation byte-identity of the vector path — I did not re-litigate.

Both findings are fixed, and the fix is wider than what was asked for: the author found
a **third** raster consumer the previous review did not name (`RsvgFfmpegBackend`, the
`attributed_rsvg` video backend at `python/termproof/cast_video.py:250`), which had the
same defect and is fixed alongside. The 320/160 pair now lives in exactly one place and
all three rasterisers read it from there, which is the same discipline the rest of the PR
argues for. That is a better fix than the one suggested.

---

## The two findings from the first review

### Finding 1 (major, raster floor lost) — **FIXED. Verified by rendering.**

`SvgRenderConfig.raster_style()` (`python/termproof/config.py:107-124`) is `style()`
floored at `RASTER_MIN_WIDTH`/`RASTER_MIN_HEIGHT`, applied as a hard lower bound with
`max`. `python/termproof/rsvg.py:113` and `python/termproof/cast_video.py:250` both call
it; `PngRenderer` (`python/termproof/builtin_renderers.py:98-99`) now reads the same two
names instead of restating `320`/`160`.

I captured the markup on its way to `rsvg-convert` and measured the real PNG, at the
merge base and at the new head:

| path | grid | base `f033dcc` | reviewed head `39398b1` | current head `e9ca247` |
|---|---|---|---|---|
| `png_rsvg` | 120x40 | 1116x836 | 1220x900 | 1220x900 |
| `png_rsvg` | 20x4 | 320x160 | **220x108** | **320x160** |
| `png_rsvg` | 3x2 | 320x160 | **50x64** | **320x160** |
| `attributed_rsvg` frames | 20x4 | 320x160 | *(unmeasured — missed)* | **320x160** |
| `attributed_rsvg` frames | 3x2 | 320x160 | *(unmeasured — missed)* | **320x160** |
| `png` (PIL) | 20x4 / 3x2 | 320x160 | 320x160 | 320x160 |
| vector SVG | 3x2 | 320x160 | 50x64 | 50x64 *(intended)* |

The regression case is gone on every raster path, and the deliberate vector change is
untouched. The 120x40 raster row moving 1116x836 → 1220x900 is the declared geometry
change, not the floor.

**Mutation-tested, not just read.** Reverting both call sites to `style()`:

```
FAIL: test_png_rsvg_rasterises_a_small_grid_at_the_floor
FAIL: test_the_video_backend_rasterises_frames_at_the_floor
Ran 797 tests ... FAILED (failures=2)
```

Both new tests drive the real renderer with a fake rasteriser and read the markup that
actually reaches the converter, rather than asserting on a style object. That was the
specific weakness called out last time (`test_an_explicit_floor_still_works_for_a_caller
_that_wants_one` pinned an unreached branch); it is genuinely addressed.

### Finding 2 (minor, the pin recipe was not exact) — **FIXED. Verified by following it literally.**

The author took the stronger of the two options: `min_width`/`min_height` are now real
`evidence.svg` keys defaulting to `0`, so the recipe can be exact rather than merely
honest about being inexact. I copied the CHANGELOG recipe byte for byte into
`.termproof/config.yaml`, loaded it through the public config loader, and compared
rendered output against the base commit:

| grid | base, unpinned | current head + recipe (with floor keys) | byte-identical? |
|---|---|---|---|
| 120x40 | 1116x836, sha `f0b956c1c99d3ca8` | 1116x836, sha `f0b956c1c99d3ca8` | **yes** |
| 3x2 | 320x160, sha `19aef86ca13088e6` | 320x160, sha `19aef86ca13088e6` | **yes** |

Same shas, not just same dimensions. Without the two floor keys, 3x2 gives 63x76 — which
is exactly what the CHANGELOG now warns about instead of claiming otherwise.

The boundary the CHANGELOG states — "narrower than 32 columns or shorter than 7 rows" —
is arithmetically exact. I checked it: at the old metrics, width `cols*9+36` is below 320
up to and including 31 columns and clears it at 32; height `rows*20+36` is below 160 up
to and including 6 rows and clears it at 7.

The new keys are also properly wired into validation. Verified by running:

```
min_width: 320 / min_height: 160   ACCEPTED
min_width: -1                      REJECTED: evidence.svg.min_width must be at least 0, got -1
min_width: 1.5                     REJECTED: evidence.svg.min_width must be int, got 1.5
bogus_key: 1                       REJECTED: unknown evidence.svg option(s) ['bogus_key'];
                                   known options: [... 'min_height', 'min_width' ...]
```

and `BUILTIN_DEFAULTS["evidence"]["svg"]` picks the two up automatically, because it is
still derived with `asdict()` rather than restated — the property the PR exists to
establish still holds after the addition.

## The other two minor findings

### Finding 3 (font stack absence assertion) — **consciously left, with the cost named.**

Acceptable, and I think correctly decided. The reason given is the coupling the previous
review flagged: Rust's `FONT_STACK` is the same literal string and the two
implementations emitting byte-identical SVG is a property the project asserts, so
appending the off-Linux names is a two-language change. That reasoning is sound and the
cost is now written down — off-Linux viewers get generic `monospace` where they
previously got SF Mono, Menlo or Consolas.

**One nit:** the cost is recorded only in the docstring of
`test_the_stack_names_nothing_that_only_exists_off_linux`
(`python/tests/test_svg_geometry_defaults.py:295-305`). The `font_family` entry in the
CHANGELOG still explains only why the *old* stack was bad on Linux; a user reading the
release notes on macOS or Windows is not told their screenshots will now render in the
browser's generic monospace. One sentence in the CHANGELOG would put it where users
look, not just where maintainers look.

### Finding 4 (`must be positive, got 0.5`) — **FIXED. Verified by running it.**

```
evidence.svg.char_width must be at least 1, got 0.5
evidence.svg.line_height must be at least 1, got 0.9
evidence.svg.padding must be at least 0, got -1
evidence.svg.font_size must be at least 1, got 0
```

Naming the bound fixes the float case and the eight int fields that had lost the word
"integer" in one move, and the type check still reports separately
(`must be int, got 1.5`).

---

## New findings from the fix round

### 5. The two new floor constants are the only geometry constants that can drift silently — minor

- **severity:** minor
- **location:** `python/termproof/config.py:50-52`, and the omission at
  `python/tests/test_svg_geometry_defaults.py:65-77`
- **verified by mutation, not inferred**

`CanonicalConstantsTest.test_each_constant_is_pinned` exists, in its own words, so that
"moving one is a deliberate act with a diff", and it pins all nine vector constants to
literals. `RASTER_MIN_WIDTH` and `RASTER_MIN_HEIGHT` are not in that list, and every new
raster test asserts against the constant rather than against `320`/`160` — so those
assertions are satisfied by whatever the constant happens to say.

**Concrete failure case, executed:**

```
RASTER_MIN_WIDTH  = 320 -> 300 : Ran 797 tests ... OK
RASTER_MIN_HEIGHT = 160 -> 130 : Ran 797 tests ... OK
```

Both mutations pass the entire suite. (A larger move is caught incidentally — 320 → 200
fails 3 tests — but only because it drops below the natural width of the 20x4 grid the
tests happen to use, which is luck rather than coverage.) The existing literal `320` in
`test_png_renderer.py:23` and the longhand reimplementation in `test_evidence_config.py:
150` do not bind either, because their grids are naturally wider than the floor.

This is precisely the drift class the whole PR is written to eliminate, now reintroduced
one level down in the new constants.

**What I would do:** add `self.assertEqual(320, RASTER_MIN_WIDTH)` and
`self.assertEqual(160, RASTER_MIN_HEIGHT)` to `test_each_constant_is_pinned`. Two lines.

### 6. The new keys are documented as raising the floor for `png`, and they do not — minor

- **severity:** minor
- **location:** `python/README.md:226`, and the same claim at `CHANGELOG.md:115-118`
- **verified by running it**

The README bullet reads: "`png`, `png_rsvg` and the `attributed_rsvg` video backend
always render at least 320x160 … Setting them higher raises that floor; setting them
lower does not lower it." The second half is true of the two rsvg paths and false of
`png`: `PngRenderer` is configured by `evidence.png`, never sees `evidence.svg`, and
ignores the keys entirely.

**Concrete failure case:** with `evidence.svg.min_width: 800` / `min_height: 600` in
`.termproof/config.yaml`, on a 20x4 grid —

```
png_rsvg -> 800x600
png      -> 320x160
```

A user who wants a larger floor on their PNG screenshots follows the documented knob and
nothing happens, with no error. The CHANGELOG's "On a raster renderer they raise the
floor but cannot lower it below 320x160" has the same gap.

**What I would do:** one clause — say the two keys raise the floor on the SVG-derived
raster paths only, and that `png` takes its geometry from `evidence.png`. Or, if the knob
is meant to be uniform, that is a separate change.

### 7. Rebase needed before merge — minor, mechanical

`CHANGELOG.md` conflicts with the current default branch (see STALENESS). Both sides
replace `Nothing yet.` under `[Unreleased]`; keeping both entry sets is the whole
resolution. I verified the resolved tree: 811 tests OK, lint and types clean. Flagging it
only because the green CI badge on this PR was produced against a base that has since
moved twice.

---

## What is good in this round

**Finding the third consumer.** `RsvgFfmpegBackend` had the identical defect and the
first review missed it — a small cast would have become a small video, not just a small
screenshot. The author looked for the class of bug rather than patching the instance
reported.

**The fix is applied at the right altitude.** `raster_style()` is one method, on the type
that already owns the geometry, and the floor value is one pair of named constants that
`PngRenderer` also now reads. The alternative — flooring inside each renderer — would
have left three copies of two numbers, which is the bug this PR exists to remove. The
distinction between vector and raster is now expressed in the type, not in a comment.

**The tests test the renderers, not the config object.** Both rsvg tests drive the real
class with a fake converter and assert on the markup that actually leaves the process,
and the PIL test opens the resulting PNG and reads its pixel size. Reverting either
production call site fails them; removing the `min_width`/`min_height` propagation from
`style()` fails four tests, including the reachability test at both fields.

**The reachability test is now genuinely complete.** `test_every_style_field_is_reachable
_from_the_render_config` exempts only `columns`/`rows`, which are the grid rather than a
configured value. Every other `SvgStyle` field must be settable from YAML or the test
fails — which answers the "no field can drift silently" question for the floors, even
though the *values* of the raster constants still can (finding 5).

**The self-correction about Rust is accurate, and I checked it rather than taking it.**
`rust/crates/termproof/src/evidence/screenshot.rs:178-181` and
`.../cast_video.rs:193-200` both invoke the converter with `--output <png> <svg>` and no
`-w`/`-h`/`-z`, and there is no occurrence of `min_width` anywhere under `rust/`. So the
Rust raster path has never been floored, Python's has, the two already disagreed before
this PR, and restoring Python's floor restores that status quo rather than widening the
gap. The CHANGELOG now says exactly that and defers the Rust-side floor deliberately.
Correcting your own changelog against your own evidence is the behaviour you want.

**Nothing regressed that the previous review had cleared.** The corpus is untouched by
this commit (the delta contains no artifact files, and the byte-identity gate passes),
the Rust tree is still empty of changes, the vector defaults are unchanged, `screen_svg`
with a default style still produces the same bytes as before the whole PR, and adding two
fields to the end of a frozen dataclass leaves positional construction intact.

---

## Not decided here

The version question raised in the first review is unchanged and is the maintainer's:
the change is marked breaking and alters every rendered artifact, the CHANGELOG's own
`0.x` rule points at a minor bump, the version is still `0.4.0` in all three files, and
`check_version.py` checks agreement between those files rather than the rule. I re-ran it
at the current head — it still reports consistent. Nothing in this fix round changes that
picture, and it did not affect my verdict on the code.

## What I could not verify

- No `rsvg-convert` is available to me, so every rsvg measurement is the SVG's
  `width`/`height` attributes on the markup handed to the converter, not pixels read back
  out of a rasterised PNG. That is a faithful proxy — the converter is invoked with no
  `-w`/`-h`/`-z`, which I confirmed in both the Python and Rust call sites — but it is a
  proxy. The PIL `png` path *was* measured as real pixels.
- I did not re-run the container images, so the font-resolution reasoning is unchanged
  from the first review and still carries the same "inferred, not measured" label there.

md-mt:
Rebased and both nits fixed in `29b141b`. Head is now `29b141b`, rebased onto `fc8adbf`.

## Finding 7 — rebased

`origin/main` had moved twice (#192, #193). `CHANGELOG.md` was the only conflict, exactly as you called it: both sides replaced `Nothing yet.` under `[Unreleased]`. Both entry sets are kept, merged under the single `### Python — Changed` heading the file's convention calls for — the docs entry from #192, then #193's dedup entry, then this PR's five, then `### Rust — Changed`. Nothing dropped from either side.

811 tests OK on the resolved tree, matching your number.

## Finding 6 — fixed, and it reproduced exactly

```
evidence.svg.min_width=800 / min_height=600, 20x4 grid:
  png_rsvg -> 800x600
  png      -> 320x160
```

You are right, and the wording was doing real harm: it pointed a user at a knob that silently does nothing for the renderer they are most likely using. Both the README bullet and the CHANGELOG entry now say the keys reach the two SVG-derived raster paths only, that `png` is configured by `evidence.png` and keeps its own fixed floor, and that making one knob govern both means giving `PngRenderConfig` its own pair — a separate change, as you say.

## Finding 5 — taking it

My call, and the answer is obviously yes: this PR exists to eliminate exactly this, and I reintroduced it one level down in the new constants. Your diagnosis of *why* the suite missed it is the part I want to acknowledge — every raster test asserts against the names, so those assertions are satisfied by whatever the constant happens to say, and the `320 → 200` case only failing because it drops below the natural width of the 20x4 grid is luck, not coverage. Now pinned in `test_each_constant_is_pinned` alongside the nine vector constants, with a comment saying why they belong in that list. Your exact mutation:

```
RASTER_MIN_WIDTH = 320 -> 300 : Ran 18 tests ... FAILED (failures=1)
```

## One thing worth recording about the mutation check

When I first ran your `320 → 300` mutation and restored it, the suite kept failing against a file that read `320`. `300` and `320` are the same byte length, and `cp` restored within one mtime tick, so CPython's `(mtime, size)` bytecode check reused the stale `.pyc`. Clearing `__pycache__` resolved it. Not a code issue and nothing changed for it — noting it because it is a trap for anyone else mutation-testing numeric constants in this repo, and it produces a failure that looks like a real one.

## Verification at `29b141b`

- `unittest discover -s tests` → **811 passed** (797 pre-rebase, +14 from #192/#193).
- `ruff` clean, `mypy termproof` clean (42 files, +1 from #193's `termproof/dedup.py`).
- `check_version.py` and `check_schema_copies.py` pass.
- `git diff origin/main -- rust/` still empty.

Version is still `0.4.0`. Your reading that the stated `0.x` rule points at `0.5.0` is on the record in both reviews and remains the captain's call — I have not touched it, and the hold still could not be registered from here for the same sandbox reason you hit.

PR #196, opened 2026-08-20
evidence: record_session, the five-step whole-session recording, in both implementations

md-mt:
## What this is

`EvidenceCollector.record_session`, in both implementations, plus the
`EvidencePublisher` video-converter seam it needs. This is the wiring 0.4.0 left
out: `Recording`, `attach_recording`, the cast-to-video converters and
`ArtifactUploader` all existed, and nothing joined them, so a downstream
consumer still carried `publish_media(config, save_cast)` and got the error
paths wrong twice before it settled.

The five steps, in order:

1. write the live session's cast, through a caller-supplied closure/callable
2. append the captured checkpoints to it, through the existing
   `append_checkpoint_frames`
3. convert the cast to a video, through the publisher's new video-converter seam
4. upload the video, through the publisher's existing uploader
5. attach a `Recording` describing what came of all that, which `publish` writes
   into the manifest

## The error-path matrix

**No failure fails the run.** Rust returns nothing rather than a `Result`;
Python raises nothing. A recording is evidence *about* a run, not part of its
verdict.

**Every failure names its step.** The error is prefixed with one of four strings
— `save cast`, `append checkpoint frames`, `convert`, `upload` — spelled once
per implementation as a constant, asserted literally in the tests, and compared
across the two by the conformance corpus. Two failures in one call are joined
with `"; "`, so `Recording.error` staying one field costs nothing.

**Which later steps run is a stated rule, not control flow:** *a step runs only
when the step before it produced the thing it works on.*

| step that failed | what still runs | why |
|---|---|---|
| 1, save cast | nothing | no cast to append to, convert or upload |
| 2, append frames | 3 and 4 | the cast is on disk and still convertible; it just ends on the session instead of the evidence |
| 3, convert | nothing after it | there is no video to upload |
| 4, upload | — | it is last; `video` is still recorded, only `url` is missing |

The case worth naming: **an upload is never attempted after a failed
conversion.** Uploading the path a conversion did not produce is how a run ends
up reporting a store error for a video that was never encoded.

Two more decisions, both tested:

- A `save_cast` that reports success and writes nothing is **step 1** failing,
  not step 2 finding a missing file. The step that lied is the one worth naming.
- A publisher with no *uploader* is not a failure — it was not asked to upload,
  exactly as in `publish`. A publisher with no *converter* is, because
  converting is what `record_session` was called to do.

Step 5 is the only step that cannot fail: it appends to a list.

## The seam

`EvidencePublisher` gains `video_converter`, beside `renderer` and `uploader`,
in the builder style already there (`with_video_converter` in Rust, a keyword
field in Python). It is `Option`/`| None` rather than defaulted: a converter is
two more binaries on the host, and a publisher that only writes stills should
not imply them.

Rust holds a concrete `CastVideoConverter`, mirroring the concrete
`ScreenshotRenderer` beside it. Python declares a `VideoConverterLike` protocol
and `RsvgFfmpegBackend` gains `convert` / `output_path_for` to satisfy it —
`render` is the `VideoBackend` protocol method and is untouched; `convert`
supplies the two things it makes a caller invent, an output path and a frame
rate.

## Tests

**Per failure path, in both languages.** Thirteen new Rust tests and fourteen new
Python ones: happy path, save refusing, save lying, append failing, no
converter, conversion failing (and the upload that must not follow), upload
raising, upload declining silently, no uploader, two failures joined, repeated
labels not sharing a cast, and `record_session` sitting beside
`attach_recording`.

**Mutation check.** 13 Rust mutants and 16 Python ones, one per behaviour the
new tests claim to guard, applied and reverted mechanically. First pass: all 13
Rust killed, 14/16 Python killed. The two survivors were real gaps — keeping
only the first of two failures, and clearing the video when an uploader declines
without a reason — and are fixed in the second commit, after which all 29 die.
Three of the Rust mutations were also run against the differential harness
alone, which killed all three. Nothing stayed green.

**Conformance.** `probe_evidence_manifest.py` and
`differential_evidence_manifest.rs` now drive `record_session` once per outcome
— `recorded`, `unsaveable`, `silent-save`, `no-converter`, `bad-encode`,
`no-url` — over publishers that differ only in the seam under test, and the
corpus is regenerated. The error strings are compared, not merely claimed to
match; so are the casts and videos the recordings write into the publish
directory. The corpus reproduces byte-for-byte from a clean run, and the Rust
half passed against it first time.

One failure is deliberately **not** in the corpus: an append that fails. Its
message comes from `append_checkpoint_frames`, which words its complaints
differently in the two languages (`empty cast file` against `<path> is empty: a
cast has a header line`), so recording it would freeze a pre-existing divergence
rather than check a shared surface. Each side covers it in its own unit tests,
and `conformance/README.md` says so.

## Deliberate differences between the two

- **`save_cast` reports failure by `Err(String)` in Rust and by raising in
  Python** — the house pattern. The message that reaches `Recording.error` is
  the same either way.
- **An uploader's own reason.** Rust reads `ArtifactUploader::last_error()` when
  it has one; Python's `UploaderLike` has no such accessor, so a Python uploader
  says why by raising. Both fall back on the same `uploader returned no URL`
  when there is no reason to be had, which is what the corpus records. Widening
  the Python protocol would break every existing uploader and is not this
  change.
- **One test differs in strength.** Python's
  `an_append_failure_does_not_stop_the_conversion` asserts the video survived,
  because its stub converter does not read the cast. Rust's real
  `CastVideoConverter` cannot read the empty cast the test uses either, so its
  version asserts step 3 was *attempted* and both failures were recorded —
  which is the same rule, and Python covers the join separately in
  `two_failures_are_both_recorded`.

## Validation

- Rust: 549 tests across 19 binaries, 0 failures; `clippy --all-targets` clean;
  `cargo fmt --check` clean.
- Python: 841 tests, 0 failures; `run_stdlib_tests.py` (149) green, so
  `cast_video` still holds its standard-library-only claim; `ruff check` and
  `mypy` clean.
- `conformance/probe_evidence_manifest.py` reproduces the committed corpus
  byte-for-byte.
- No version bump. The changelog entries are under `[Unreleased]`.

## Noticed and deliberately left alone

Out of scope, noted rather than fixed:

- **`append_checkpoint_frames` words the same two failures differently in the
  two languages** (`empty cast file` against `<path> is empty: a cast has a
  header line`). It predates this change and is why that one failure stays out
  of the conformance corpus. Worth converging, in its own diff.
- **The two `VideoBackend` shapes.** Rust has `evidence::video::VideoBackend`
  (`render(&Path, &Path, u32) -> Result<(), VideoError>`) *and* the concrete
  `CastVideoConverter`; Python has `protocols.VideoBackend` and now
  `collector.VideoConverterLike`. Four names for two ideas. Unifying them is a
  publisher refactor.
- **`store::sanitize_component` returns `"default"` for an empty label where
  Python's `_sanitize` returns `""`.** Pre-existing, reachable by both
  `CapturedStep::file_stem` and the new recording stem, and not exercised by any
  caller in tree.
- **`RsvgFfmpegBackend.convert` defaults to `DEFAULT_FPS` (24) while
  `VideoConfig.fps` defaults to 60.** `convert` takes the configured rate when
  the backend was built `from_config`, and 24 — Rust's `CastVideoConverter`
  default — otherwise. Reconciling the two written-down defaults is a config
  change with its own compatibility question.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -258,39 +258,6 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   "must be a positive integer" was wrong for them, and `0.5` satisfies
   "positive" while still being refused.
 
-### Rust — Changed (breaking)
-
-- **`evidence::collector::EvidencePublisher` gained a fifth public field,
-  `video_converter: Option<CastVideoConverter>`, which stops the struct being
-  constructible by literal.** Every field of `EvidencePublisher` is `pub` and
-  the struct is not `#[non_exhaustive]`, so an external crate writing
-  `EvidencePublisher { dir, identity, renderer, uploader }` no longer compiles;
-  `cargo semver-checks` reports it as `constructible_struct_adds_field`. Nothing
-  else about the type moved — the four existing fields keep their names, types
-  and meanings, and `EvidencePublisher::new` plus the `with_*` builders
-  (`with_renderer`, `with_uploader`, and the new `with_video_converter`) compile
-  unchanged.
-
-  **To fix a literal, build through the constructor**, which is what every
-  in-tree caller and every doc example already does:
-
-  ```rust
-  // was
-  let publisher = EvidencePublisher { dir, identity, renderer, uploader: Some(u) };
-  // now
-  let publisher = EvidencePublisher::new(dir, identity)
-      .with_renderer(renderer)
-      .with_uploader(u);
-  ```
-
-  Filed as breaking rather than as an addition because the pre-1.0 rule in this
-  file's preamble is about what a consumer's source does, not about what the
-  crate meant to offer. The seam had to hang somewhere, and the alternative
-  shape — passing the converter to `record_session` instead of the publisher —
-  would have avoided the break at the cost of putting one of the publisher's
-  three seams somewhere other than the publisher. That trade is worth stating;
-  it is not free either way.
-
 ### Rust — Added
 
 - **`result::score_from` and `result::assertion_map`:** the same two functions
@@ -349,6 +316,39 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   status quo rather than changing it. Giving `SvgMetrics` a raster floor to
   match is worth doing and is deliberately not done here — it would change what
   Rust renders, in a release that changes no other Rust rendering behaviour.
+
+### Rust — Changed (breaking)
+
+- **`evidence::collector::EvidencePublisher` gained a fifth public field,
+  `video_converter: Option<CastVideoConverter>`, which stops the struct being
+  constructible by literal.** Every field of `EvidencePublisher` is `pub` and
+  the struct is not `#[non_exhaustive]`, so an external crate writing
+  `EvidencePublisher { dir, identity, renderer, uploader }` no longer compiles;
+  `cargo semver-checks` reports it as `constructible_struct_adds_field`. Nothing
+  else about the type moved — the four existing fields keep their names, types
+  and meanings, and `EvidencePublisher::new` plus the `with_*` builders
+  (`with_renderer`, `with_uploader`, and the new `with_video_converter`) compile
+  unchanged.
+
+  **To fix a literal, build through the constructor**, which is what every
+  in-tree caller and every doc example already does:
+
+  ```rust
+  // was
+  let publisher = EvidencePublisher { dir, identity, renderer, uploader: Some(u) };
+  // now
+  let publisher = EvidencePublisher::new(dir, identity)
+      .with_renderer(renderer)
+      .with_uploader(u);
+  ```
+
+  Filed as breaking rather than as an addition because the pre-1.0 rule in this
+  file's preamble is about what a consumer's source does, not about what the
+  crate meant to offer. The seam had to hang somewhere, and the alternative
+  shape — passing the converter to `record_session` instead of the publisher —
+  would have avoided the break at the cost of putting one of the publisher's
+  three seams somewhere other than the publisher. That trade is worth stating;
+  it is not free either way.
 
 ## [0.4.0] — 2026-08-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,12 +87,23 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   one non-fatal step; **a conversion that failed leaves nothing to upload, so no
   upload is attempted** — reporting a store error for a video that was never
   encoded is exactly what two independent consumer implementations got wrong.
-  Nothing raises: a recording is evidence about a run, not part of its verdict.
-  Every failure instead lands on the `Recording`'s `error`, prefixed with the
-  name of the step that produced it — `save cast`, `append checkpoint frames`,
-  `convert` or `upload` — and two failures in one call are joined with `"; "` so
-  neither is lost. A callable that reports success and writes no cast is
-  reported as step 1 failing, not left for step 2 to discover as a missing file.
+  No failure a step *reports* raises: a recording is evidence about a run, not
+  part of its verdict. Every failure instead lands on the `Recording`'s `error`,
+  prefixed with the name of the step that produced it — `save cast`,
+  `append checkpoint frames`, `convert` or `upload` — and two failures in one
+  call are joined with `"; "` so neither is lost. The promise stops where
+  `Exception` does: a `KeyboardInterrupt` or `SystemExit` from one of the three
+  caller-supplied seams propagates and the recording is lost rather than
+  degraded, which the docstring now says outright.
+
+  **A step is not believed, it is checked.** Each of the three seams is
+  caller-supplied, and the outcome worse than any recorded failure is a
+  `Recording` that reports none — no `error`, a `video` that is not on disk, and
+  a `url` for it. So a `save_cast` that returns having written no file, a
+  converter that returns a path it did not write, and an uploader that returns
+  an empty string are each recorded as *that* step failing
+  (`convert: reported success but wrote no file at <path>`, and so on), rather
+  than left for the next step to trip over and take the blame.
 
 - **`termproof.collector.EvidencePublisher.video_converter`**, beside the
   existing `renderer` and `uploader`, with `termproof.collector.VideoConverterLike`
@@ -247,6 +258,39 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   "must be a positive integer" was wrong for them, and `0.5` satisfies
   "positive" while still being refused.
 
+### Rust — Changed (breaking)
+
+- **`evidence::collector::EvidencePublisher` gained a fifth public field,
+  `video_converter: Option<CastVideoConverter>`, which stops the struct being
+  constructible by literal.** Every field of `EvidencePublisher` is `pub` and
+  the struct is not `#[non_exhaustive]`, so an external crate writing
+  `EvidencePublisher { dir, identity, renderer, uploader }` no longer compiles;
+  `cargo semver-checks` reports it as `constructible_struct_adds_field`. Nothing
+  else about the type moved — the four existing fields keep their names, types
+  and meanings, and `EvidencePublisher::new` plus the `with_*` builders
+  (`with_renderer`, `with_uploader`, and the new `with_video_converter`) compile
+  unchanged.
+
+  **To fix a literal, build through the constructor**, which is what every
+  in-tree caller and every doc example already does:
+
+  ```rust
+  // was
+  let publisher = EvidencePublisher { dir, identity, renderer, uploader: Some(u) };
+  // now
+  let publisher = EvidencePublisher::new(dir, identity)
+      .with_renderer(renderer)
+      .with_uploader(u);
+  ```
+
+  Filed as breaking rather than as an addition because the pre-1.0 rule in this
+  file's preamble is about what a consumer's source does, not about what the
+  crate meant to offer. The seam had to hang somewhere, and the alternative
+  shape — passing the converter to `record_session` instead of the publisher —
+  would have avoided the break at the cost of putting one of the publisher's
+  three seams somewhere other than the publisher. That trade is worth stating;
+  it is not free either way.
+
 ### Rust — Added
 
 - **`result::score_from` and `result::assertion_map`:** the same two functions
@@ -259,15 +303,15 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   steps with the same semantics and the same rule about which of them run after
   a failure, taking `save_cast: FnOnce(&Path) -> Result<(), String>` where
   Python takes a callable that raises. It returns nothing rather than a
-  `Result`, for the reason nothing on the Python side raises: no failure may
-  fail the run. `&mut self`, because it appends a `Recording`; `publish` stays
-  `&self`. The four step names it prefixes an error with are the same four
-  strings both implementations write, which the conformance pair now compares.
+  `Result`, for the reason nothing on the Python side raises: no failure a step
+  reports may fail the run. `&mut self`, because it appends a `Recording`;
+  `publish` stays `&self`. The four step names it prefixes an error with are the
+  same four strings both implementations write, which the conformance pair now
+  compares.
 
-- **`evidence::collector::EvidencePublisher::video_converter`** and
-  `with_video_converter`, beside the existing `renderer` and `uploader` and in
-  the same builder style. Holds a `CastVideoConverter`, and is `Option` for the
-  reason the Python field is.
+- `with_video_converter`, beside the existing `with_renderer` and
+  `with_uploader` and in the same builder style. The field it sets is the
+  breaking part above.
 
 - **`evidence::cast_video::append_checkpoint_frames`:** the same capability with
   the same semantics, taking `hold_seconds: Option<f64>` where Python takes a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,44 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   same rule; the difference between the two shapes is that a list weighs
   duplicate names once each and a mapping folds them into one entry.
 
+- **`termproof.collector.EvidenceCollector.record_session`:** the wiring 0.4.0
+  left out. `Recording`, `attach_recording`, the cast-to-video backends and the
+  upload seam all existed; nothing joined them, so every consumer that wanted a
+  video of a whole run wrote the same five-step sequence itself. This is that
+  sequence: save the live session's cast through a caller-supplied callable,
+  append the captured checkpoints to it with `append_checkpoint_frames`,
+  convert it to a video, upload the video, and attach a `Recording` that
+  `publish` then writes into the manifest.
+
+  **The error handling is the part worth having upstream, and it is a stated
+  rule rather than an accident of control flow: a step runs only when the step
+  before it produced the thing it works on.** A cast that could not be saved
+  leaves nothing to append to, convert or upload, so steps 2–4 are skipped; an
+  append that failed leaves the cast on disk and still convertible, so it is the
+  one non-fatal step; **a conversion that failed leaves nothing to upload, so no
+  upload is attempted** — reporting a store error for a video that was never
+  encoded is exactly what two independent consumer implementations got wrong.
+  Nothing raises: a recording is evidence about a run, not part of its verdict.
+  Every failure instead lands on the `Recording`'s `error`, prefixed with the
+  name of the step that produced it — `save cast`, `append checkpoint frames`,
+  `convert` or `upload` — and two failures in one call are joined with `"; "` so
+  neither is lost. A callable that reports success and writes no cast is
+  reported as step 1 failing, not left for step 2 to discover as a missing file.
+
+- **`termproof.collector.EvidencePublisher.video_converter`**, beside the
+  existing `renderer` and `uploader`, with `termproof.collector.VideoConverterLike`
+  as its protocol. Optional rather than defaulted, because a converter is two
+  more binaries on the host; a `record_session` against a publisher without one
+  records that as the `convert` step failing, since converting is what it was
+  called to do.
+
+- **`termproof.cast_video.RsvgFfmpegBackend.convert`** and `output_path_for`,
+  which satisfy that protocol: `render` is the `VideoBackend` method and is
+  unchanged, and `convert` supplies the two things it makes a caller invent — an
+  output path and a frame rate. The rate is the configured
+  `evidence.video.fps`, or `DEFAULT_FPS` when the backend was not built from a
+  config, which is the rate Rust's `CastVideoConverter` defaults to.
+
 - **`termproof.cast_video.append_checkpoint_frames`:** appends the captured
   checkpoint screens to a cast as held trailing frames, so a recording ends by
   replaying the evidence sequence instead of stopping on whatever the last
@@ -216,6 +254,20 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   `assertion_map` collecting `(impl Into<String>, bool)` pairs. The empty-map
   case is `1.0` here too, and is documented on `RunResult::score` — the field
   the number ends up in — rather than only on the function that computes it.
+
+- **`evidence::collector::EvidenceCollector::record_session`:** the same five
+  steps with the same semantics and the same rule about which of them run after
+  a failure, taking `save_cast: FnOnce(&Path) -> Result<(), String>` where
+  Python takes a callable that raises. It returns nothing rather than a
+  `Result`, for the reason nothing on the Python side raises: no failure may
+  fail the run. `&mut self`, because it appends a `Recording`; `publish` stays
+  `&self`. The four step names it prefixes an error with are the same four
+  strings both implementations write, which the conformance pair now compares.
+
+- **`evidence::collector::EvidencePublisher::video_converter`** and
+  `with_video_converter`, beside the existing `renderer` and `uploader` and in
+  the same builder style. Holds a `CastVideoConverter`, and is `Option` for the
+  reason the Python field is.
 
 - **`evidence::cast_video::append_checkpoint_frames`:** the same capability with
   the same semantics, taking `hold_seconds: Option<f64>` where Python takes a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -350,6 +350,27 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   three seams somewhere other than the publisher. That trade is worth stating;
   it is not free either way.
 
+  **The version does not move for it.** The pre-1.0 rule above says a breaking
+  change bumps the minor digit; the decision on [#196](https://github.com/md-mt/termproof/pull/196)
+  is that termproof stays on 0.4.x and this break is waived instead. So
+  `constructible_struct_adds_field` is set to `allow` in
+  `[package.metadata.cargo-semver-checks.lints]` in
+  `rust/crates/termproof/Cargo.toml` — the tool's own scoped mechanism, one
+  lint, rather than a version bump or a disabled CI job. **This entry is not
+  rewritten to match:** a waiver decides what the version does about a break,
+  not whether the break happened, and a consumer whose literal stopped
+  compiling needs the paragraphs above whatever the digit says.
+
+  The waiver is scoped in two directions, because at lint granularity it would
+  otherwise also cover field additions nobody has decided about:
+  `every_public_field_of_the_publisher_is_accounted_for` builds
+  `EvidencePublisher` from an exhaustive literal, so a sixth field fails to
+  compile, and
+  `the_semver_waiver_is_scoped_to_the_release_it_was_granted_for` fails the
+  build if the version leaves 0.4.x while the waiver is still in place. The
+  release checklist in `rust/docs/publishing.md` now says to read the waiver
+  list before reading the green check.
+
 ## [0.4.0] — 2026-08-19
 
 ### Rust — Changed
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -432,15 +432,44 @@ base of `0.5000005` — which is what a Rust-recorded cast ends on, since
 `CastRecorder` writes `as_secs_f64()` unrounded — separates them at three of the
 eight appended frames.
 
+The same reasoning brought `record_session` in. Its outcome is a string in the
+manifest — the `error` field of a `Recording`, prefixed with the name of the
+step that failed — and "which step failed" is only useful to a reader if it does
+not depend on which implementation produced the run. Each half drives it once
+per outcome, over publishers that differ only in the seam under test:
+
+| recording | what it pins |
+|---|---|
+| `recorded` | all five steps succeed: the cast, the evidence appended to it, the video, the URL |
+| `unsaveable` | step 1 refusing — Python raises, Rust returns `Err`, both write `save cast: disk on fire` |
+| `silent-save` | step 1 reporting success and writing nothing, which is still step 1 |
+| `no-converter` | a publisher with no video converter, which is the `convert` step failing |
+| `bad-encode` | a conversion that fails, and the upload that must not follow it |
+| `no-url` | an upload that declines, which costs the URL and not the video |
+
+The recordings' own casts and videos land in the publish directory, so they are
+compared as files too, not only as paths.
+
+One failure is deliberately absent: an append that fails. Its message comes from
+`append_checkpoint_frames`, which words its complaints differently in the two
+languages — `empty cast file` against `<path> is empty: a cast has a header
+line` — so recording it here would freeze a divergence in the corpus rather than
+check a shared surface. Each implementation covers it in its own unit tests.
+
 ## What is deliberately neutralised
 
-Two things in the scenario are properties of the machine rather than of either
+Three things in the scenario are properties of the machine rather than of either
 implementation, and both halves handle them the same way:
 
 - **The rasteriser.** `ScreenshotRenderer` shells out to `rsvg-convert`, so on a
   host without it every step would record a render error whose text belongs to
   the operating system. Both halves stub the tool and write a fixed byte, so a
   screenshot is recorded on every host.
+- **The video encoder.** Rust's `CastVideoConverter` shells out to
+  `rsvg-convert` and `ffmpeg`; the Rust half stubs the tool runner and the
+  Python half stands in a converter that writes the same fixed byte. Neither the
+  frame count nor the encoder reaches the manifest — the video path and the file
+  at it are what have to agree.
 - **The publish directory**, which is a fresh temporary directory. Both halves
   substitute it for `@DIR` before comparing.
 
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -445,16 +445,41 @@ per outcome, over publishers that differ only in the seam under test:
 | `silent-save` | step 1 reporting success and writing nothing, which is still step 1 |
 | `no-converter` | a publisher with no video converter, which is the `convert` step failing |
 | `bad-encode` | a conversion that fails, and the upload that must not follow it |
+| `silent-encode` | step 3 reporting success and writing nothing — the `silent-save` guard, one step down |
 | `no-url` | an upload that declines, which costs the URL and not the video |
+| `blank-url` | an upload that returns `""`, which both sides have to reject rather than record |
+| *(empty label)* | the one input on which the two filename schemes could part |
 
 The recordings' own casts and videos land in the publish directory, so they are
 compared as files too, not only as paths.
 
-One failure is deliberately absent: an append that fails. Its message comes from
-`append_checkpoint_frames`, which words its complaints differently in the two
-languages — `empty cast file` against `<path> is empty: a cast has a header
-line` — so recording it here would freeze a divergence in the corpus rather than
-check a shared surface. Each implementation covers it in its own unit tests.
+The empty-label case is there because a recording's `cast` and `video` are
+strings in the manifest and the two implementations build them differently at
+exactly one input: Rust goes through `store::sanitize_component`, which
+substitutes `default` for a component that sanitises to nothing, and Python's
+`_sanitize` returns `""`. `_recording_file_stem` applies the same fallback so
+the two agree, and this case is what holds it there. `CapturedStep.file_stem`
+still differs the same way for the same input — pre-existing, not driven by this
+scenario, and listed under "outside" below.
+
+Two failures are deliberately absent:
+
+- **An append that fails.** Its message comes from `append_checkpoint_frames`,
+  which words its complaints differently in the two languages — `empty cast
+  file` against `<path> is empty: a cast has a header line` — so recording it
+  here would freeze a divergence in the corpus rather than check a shared
+  surface. Each implementation covers it in its own unit tests.
+- **An upload that declines *with a reason*.** Rust reads
+  `ArtifactUploader::last_error()`; Python's `UploaderLike` has no such
+  accessor, so a Python uploader says why by raising. `no-url` and `blank-url`
+  therefore pin the *shared fallback* — `upload: uploader returned no URL`,
+  emitted when there is no reason to be had — and not the shipped behaviour:
+  `FallbackUploader`,
+  the only non-test `ArtifactUploader` in the crate, sets `last_error` on every
+  failure, so in practice Rust writes `upload: all uploaders returned no URL`
+  where Python writes `upload: uploader returned no URL`. That disagreement is
+  real and is recorded here rather than in the corpus, because pinning it would
+  make the harness certify a divergence.
 
 ## What is deliberately neutralised
 
--- a/conformance/corpus/evidence_manifest.expected.json
+++ b/conformance/corpus/evidence_manifest.expected.json
@@ -1,5 +1,11 @@
 {
   "files": {
+    "recording-02-recorded.cast": "{\"version\":2,\"width\":80,\"height\":24}\n[0.5,\"o\",\"\\u001b[3;20rMENU\"]\n[1.25,\"o\",\"\\r\\nitem one\"]\n[4.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JMENU\\r\\nitem one\"]\n[7.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JMENU\\r\\nitem one\"]\n[10.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JRECOVERED\"]\n[13.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JLAST SCREEN\"]\n[16.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JERROR\"]\n[19.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2J\\u001b[31mALERT\\u001b[0m\"]\n[22.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JALERT\"]\n[25.25,\"o\",\"\\u001b[m\"]\n",
+    "recording-02-recorded.mp4": "mp4",
+    "recording-05-no-converter.cast": "{\"version\":2,\"width\":80,\"height\":24}\n[0.5,\"o\",\"\\u001b[3;20rMENU\"]\n[1.25,\"o\",\"\\r\\nitem one\"]\n[4.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JMENU\\r\\nitem one\"]\n[7.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JMENU\\r\\nitem one\"]\n[10.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JRECOVERED\"]\n[13.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JLAST SCREEN\"]\n[16.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JERROR\"]\n[19.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2J\\u001b[31mALERT\\u001b[0m\"]\n[22.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JALERT\"]\n[25.25,\"o\",\"\\u001b[m\"]\n",
+    "recording-06-bad-encode.cast": "{\"version\":2,\"width\":80,\"height\":24}\n[0.5,\"o\",\"\\u001b[3;20rMENU\"]\n[1.25,\"o\",\"\\r\\nitem one\"]\n[4.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JMENU\\r\\nitem one\"]\n[7.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JMENU\\r\\nitem one\"]\n[10.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JRECOVERED\"]\n[13.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JLAST SCREEN\"]\n[16.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JERROR\"]\n[19.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2J\\u001b[31mALERT\\u001b[0m\"]\n[22.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JALERT\"]\n[25.25,\"o\",\"\\u001b[m\"]\n",
+    "recording-07-no-url.cast": "{\"version\":2,\"width\":80,\"height\":24}\n[0.5,\"o\",\"\\u001b[3;20rMENU\"]\n[1.25,\"o\",\"\\r\\nitem one\"]\n[4.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JMENU\\r\\nitem one\"]\n[7.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JMENU\\r\\nitem one\"]\n[10.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JRECOVERED\"]\n[13.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JLAST SCREEN\"]\n[16.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JERROR\"]\n[19.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2J\\u001b[31mALERT\\u001b[0m\"]\n[22.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JALERT\"]\n[25.25,\"o\",\"\\u001b[m\"]\n",
+    "recording-07-no-url.mp4": "mp4",
     "session-with-checkpoints.cast": "{\"version\":2,\"width\":80,\"height\":24}\n[0.5,\"o\",\"\\u001b[3;20rMENU\"]\n[1.25,\"o\",\"\\r\\nitem one\"]\n[4.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JMENU\\r\\nitem one\"]\n[7.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JMENU\\r\\nitem one\"]\n[10.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JRECOVERED\"]\n[13.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JLAST SCREEN\"]\n[16.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JERROR\"]\n[19.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2J\\u001b[31mALERT\\u001b[0m\"]\n[22.25,\"o\",\"\\u001b[m\\u001b[r\\u001b[H\\u001b[2JALERT\"]\n[25.25,\"o\",\"\\u001b[m\"]\n",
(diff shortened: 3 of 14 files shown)


md-mt:
# PR 196 — evidence: `record_session`, the five-step whole-session recording, in both implementations

**VERDICT:** CHANGES NEEDED
**MERGE RISK:** medium — the matrix has one hole that produces a *silent false success* (a `Recording` with `error: null`, a `video` path that does not exist, and a `url` for a video that was never encoded), and the public-API compatibility check is red for a real reason that needs a version decision before merge.
**CI:** 19 pass / 1 fail / 3 skip. The one failure is `Public API compatibility (cargo semver-checks)`, and it is genuine, not an artefact (see F2). Coverage is otherwise meaningful for this change: both unit suites, the stdlib-only gate, and the differential harness all exercise the new surface, and I confirmed the differential harness really does catch cross-implementation drift in it (see "What is good").
**STALENESS:** Not stale. Base is the current default-branch tip; `git merge-tree` reports a clean merge. The immediately-preceding default-branch commit is the `append_checkpoint_frames` change this builds on, and it is present.

## Summary

This wires five existing pieces into `EvidenceCollector.record_session` in both
implementations and adds the `video_converter` seam it needs. The wiring is
right, the stated ordering rule ("a step runs only when the step before it
produced the thing it works on") is actually implemented rather than merely
described, and the two implementations agree under failure in every case I
forced except one filename. The test claims in the PR body hold up: I killed
seven mutants independently and the differential harness kills three of them on
its own. Two things stop me recommending merge as-is: step 3 has no counterpart
to the "reported success but wrote no file" guard the author built for step 1,
and the semver break is real and is filed in the changelog as an addition.

### How I checked

Everything below was run against the PR head in a scratch worktree.

- Full suites: Rust `cargo test -p termproof` — 557 passed, 0 failed; clippy
  `--all-targets` and `cargo fmt --check` clean. Python `unittest discover` —
  841 passed. Stdlib-only gate — 149 passed, 6 skipped.
- Conformance: the corpus regenerates byte-for-byte from the Python oracle
  (`diff` empty), and `differential_evidence_manifest` passes against the
  committed corpus.
- I enumerated the failure matrix from the code myself and then executed it:
  a scratch probe in each language drove `record_session` through nineteen
  forced-failure shapes (closure raising / returning `Err`, closure lying,
  closure writing a directory, closure writing an empty cast, unwritable
  publish directory, cast made read-only before the append, no converter,
  converter erroring, converter binaries absent, converter lying, converter
  returning a malformed value, uploader raising, uploader declining, uploader
  declining with a reason, uploader returning an empty URL, no uploader, zero
  captured checkpoints, empty label, punctuation-only label).
- Mutation: I broke seven specific behaviours, one at a time, and recorded
  which test went red and whether the differential pair also went red.

**Every finding below says whether I verified it by running it or inferred it
from reading.**

## Findings

### F1 — major — step 3 has no counterpart to step 1's "wrote no file" guard, so a lying converter is recorded as success

**Location:** `python/termproof/collector.py:588-611`; `rust/crates/termproof/src/evidence/collector.rs:548-570`

**Verified by execution, in both implementations.**

Step 1 explicitly refuses to trust the caller-supplied closure: if it reports
success and no file appears, that is recorded as `save cast: reported success
but wrote no file at <path>`, and the reasoning is written into the docstring —
"the step that lied is the one worth naming". Step 3 takes the caller-supplied
converter entirely on trust. If `convert` returns a path and does not create the
file, `record_session` sets `recording.video` to that path, proceeds to step 4,
uploads a path that does not exist, and records the URL the uploader hands back.

Concrete failure, run in both languages with a converter whose only defect is
that it returns without writing (in Rust, a tool runner whose `ffmpeg` exits 0
and writes nothing — the plausible shape, since `CastVideoConverter::with_runner`
is public):

```
Python
    video  : recording-00-full-session.mp4  exists=False
    url    : 'https://example/v'
    error  : None
    upload calls : [('.../recording-00-full-session.mp4', False)]

Rust
    video  : recording-00-full-session.mp4  exists=false
    url    : Some("https://example/v")
    error  : None
    uploads: [(".../recording-00-full-session.mp4", false)]
```

`error: null` with a `video` and a `url` is the manifest saying the recording
succeeded. This is the inverse of the failure the PR says it exists to prevent —
the body's own words are "reporting a store error for a video that was never
encoded" — and it lands in the same place: a reader chasing a video that is not
there, with nothing in the report to say which step went wrong. The Rust unit
test `a_failed_conversion_is_never_uploaded` asserts "an upload was attempted for
a video that does not exist" must never happen; that invariant holds only while
the converter is honest about its return.

This is not a divergence — both implementations are wrong the same way — so the
conformance corpus cannot catch it either.

It is not reachable through the two in-tree converters in normal operation
(a real `ffmpeg` that cannot write its output exits non-zero, which both
implementations already report as `convert: ...` — I confirmed that with real
binaries absent: `convert: [Errno 2] No such file or directory` and
`convert: No such file or directory (os error 2)`). It matters because
`VideoConverterLike` is a *public protocol* whose whole purpose is third-party
converters, which is exactly the argument that justified the step-1 guard.

**What I would do:** mirror step 1 — after a successful `convert`, check the
returned path is a file; if not, push `convert: reported success but wrote no
file at <path>`, leave `video` unset, and skip step 4. Add a `silent-encode`
case to the conformance corpus beside `silent-save`, since the two guards are
the same guard.

### F2 — major — the semver-checks failure is a genuine public-API break, and the changelog files it as an addition

**Location:** `rust/crates/termproof/src/evidence/collector.rs:789`; `CHANGELOG.md:236`

**Verified.** The break is real, not an artefact:

- `EvidencePublisher` is a `pub struct` with every field `pub`, no
  `#[non_exhaustive]` (I grepped), and every field type is externally
  constructible (`PathBuf`, `RunIdentity::new`, `ScreenshotRenderer::new`,
  `Option<Box<dyn ArtifactUploader>>`). An external crate could therefore write
  `EvidencePublisher { dir, identity, renderer, uploader }`, and that literal no
  longer compiles. `constructible_struct_adds_field` is firing correctly.
- The same check passed on the default branch in the run immediately before this
  PR, so this change is the cause rather than inherited redness.
- Blast radius inside the repository is nil: I found no struct-literal
  construction of `EvidencePublisher` anywhere in the tree outside the module
  that defines it, and no in-tree caller of `record_session`, `attach_recording`
  or `Recording` at all. The break falls entirely on external consumers.

Whether to bump is the maintainer's call and I am not making it. What is not a
judgement call is the record: this project's changelog already has a
`### Rust — Changed (breaking)` heading (used for the 0.4.0 dependency-surface
work), the file's own preamble states the pre-1.0 rule that a breaking change
bumps the minor digit, and this change is filed only under `### Rust — Added`
with the PR body saying "No version bump". A consumer reading the released
changelog would not learn that their `EvidencePublisher` literal stopped
compiling.

Worth naming for the version discussion: the break is unavoidable *given this
shape* — any new field, public or private, breaks an exhaustively-constructible
struct — but it is avoidable with a different shape, e.g. passing the converter
to `record_session` rather than hanging it off the publisher. The author's
argument for "beside `renderer` and `uploader`" is a reasonable one; it just is
not free, and the changelog should say so.

**What I would do:** re-file the entry under a breaking heading naming the exact
symbol, and let the maintainer decide the digit.

### F3 — minor — `record_session`'s filename stem diverges between the two implementations for an empty label

**Location:** `python/termproof/collector.py:729`; `rust/crates/termproof/src/evidence/collector.rs:705`

**Verified by execution.** `record_session("", ...)` produces:

- Rust: `recording-00-default.cast` / `.mp4`
- Python: `recording-00-.cast` / `.mp4`

so the `cast` and `video` strings in the manifest differ between
implementations for the same call. (A punctuation-only label agrees:
`recording-00----` on both.)

The PR names the underlying cause in its "noticed and deliberately left alone"
section — `sanitize_component` returns `"default"` for an empty component where
Python's `_sanitize` returns `""` — and says it is "not exercised by any caller
in tree". That was true before; `record_session` is now a caller, it is a *new*
cross-implementation surface, and it is a caller the conformance harness drives.
The corpus exercises it only with non-empty labels, so the pair would not notice.

I would not block on it. But the PR's headline claim is that the two
implementations agree on what `record_session` writes into the manifest, and
this is the one input I found where they do not.

**What I would do:** either add an empty-label recording to the corpus so the
divergence is recorded as known rather than merely unseen, or state in the
docstrings that the label must be non-empty. Converging `sanitize_component`
stays a separate change, as the author says.

### F4 — minor — the only upload-failure shape the corpus pins is the one no shipped uploader produces

**Location:** `rust/crates/termproof/src/evidence/collector.rs:557-570`; `conformance/corpus/evidence_manifest.expected.json` (`no-url`)

**Inferred from reading, with the corpus behaviour verified.** The PR documents
that Rust consults `ArtifactUploader::last_error()` while Python's `UploaderLike`
has no such accessor, and says both "fall back on the same `uploader returned no
URL` when there is no reason to be had, which is what the corpus records". That
is accurate but understates it: `FallbackUploader` — the only non-test
`ArtifactUploader` implementation the crate provides — sets `last_error` on
*every* failure, either the joined store errors or `"all uploaders returned no
URL"`. It never returns `None` from `last_error()` after a failed upload.

So with the uploader the crate actually ships, Rust writes
`upload: all uploaders returned no URL` where Python writes
`upload: uploader returned no URL`. The corpus's `no-url` case pins agreement
using a Rust stub that reports no `last_error` — a configuration that does not
occur in the crate. The differential pair therefore certifies agreement on step
4 for a case that is not the practical one.

Widening the Python protocol is correctly out of scope. Recording the *disagreement*
is not.

**What I would do:** say in `conformance/README.md`, beside the existing note
about the append message, that step 4's message agrees only when the Rust
uploader has no `last_error`, and that the shipped `FallbackUploader` always
does — so `no-url` pins the shared fallback, not the shipped behaviour.

### F5 — nit — an uploader that returns an empty string is recorded as a successful upload

**Location:** `python/termproof/collector.py:606`; `rust/crates/termproof/src/evidence/collector.rs:559`

**Verified by execution, both implementations agree.** An uploader returning
`""` yields `url: ""` and `error: null`. The Rust side already treats an empty
`last_error()` as "no reason given" (`.filter(|e| !e.is_empty())`), so the same
scepticism about empty strings is one line away and is not applied to the URL.
Consistent across languages, so harmless for parity; still a recording that
claims a URL it does not have.

### F6 — nit — the Python README example calls a method that does not exist

**Location:** `python/README.md:130`

**Verified.** `collector.record_session("full-session", publisher, lambda dest: session.save_cast(dest))`
— there is no `save_cast` anywhere in the Python package (I grepped; the only
match is the `record_session` parameter itself), and `session` is undefined in
the example. A Python session exposes `session.cast_path`, so the honest one-liner
is a copy from there. The Rust doc example does exactly that and is
compile-checked as a `no_run` doctest; the Python one is prose and is not
checked, which is why it drifted.

### F7 — nit — "no failure may fail the run" has an unstated boundary

**Location:** `python/termproof/collector.py:488-611`

**Verified by execution.** Python catches `Exception`, so a `KeyboardInterrupt`
or `SystemExit` raised by any of the three caller-supplied seams propagates —
and because the `Recording` is only appended at the end of each branch, the
recording is lost entirely rather than degraded. Rust's equivalent is a panic in
the closure or the runner, which unwinds the same way. I think this is the right
behaviour and I am not asking for it to change; the docstring's absolute phrasing
("Nothing here raises, on purpose") is what I would qualify, since the method's
whole contract is about what a caller can rely on when things go wrong.

## What is good

Genuinely, and this is where I would *not* spend more review attention:

- **The ordering rule is implemented, not just documented.** I forced each step
  to fail in both languages and the run survived every time, with the recording
  attached and the step named. Upload-after-failed-conversion never happens
  (verified: zero upload calls in both). An append failure does not stop steps 3
  and 4 (verified: video and URL both present after a read-only cast made the
  append fail).
- **The mutation claim holds, and I checked it independently rather than taking
  it on the PR's word.** Seven mutants, each killed by exactly the test that
  should own it:

  | mutation | test that went red |
  |---|---|
  | delete step 1's "wrote no file" guard (Py) | `test_a_save_that_writes_nothing_is_the_save_step_failing` |
  | upload even after a failed conversion (Py) | `test_a_failed_conversion_is_never_uploaded` |
  | keep only the first of two errors (Py) | `test_two_failures_are_both_recorded` |
  | rename the `convert` step string (Py) | three tests |
  | keep only the first of two errors (Rust) | `an_append_failure_does_not_stop_the_conversion` |
  | upload even after a failed conversion (Rust) | `a_failed_conversion_is_never_uploaded` |
  | clear `video` when the upload declines (Rust) | `a_failed_upload_keeps_the_video` |

- **The conformance pair earns its place.** This was the thing most likely to be
  decorative, and it is not: mutating the Rust side alone — renaming the
  `convert` step string, uploading after a failed conversion, or clearing the
  video on a declined upload — turns `differential_evidence_manifest` red
  against the committed corpus. Mutating the Python side alone changes the
  regenerated corpus, which would turn the Rust half red in the same way. The
  two "join the errors" mutants survive the pair, which is exactly what the
  documented decision to keep the append failure out of the corpus predicts.
  The corpus also regenerates byte-for-byte from the oracle on this machine.
- **The decision to keep the append failure out of the corpus is right**, and it
  is written down in all three places a reader might look (the probe docstring,
  the harness module doc, and `conformance/README.md`). Recording it would have
  frozen a known pre-existing divergence as if it were agreed behaviour.
- **Choosing to treat a lying `save_cast` as step 1** is the subtle call in the
  whole change and it is the correct one — see F1, which is only a finding
  because the same reasoning was not carried one step further.
- **Blast radius is genuinely nil in-tree.** No production code calls
  `record_session`, `attach_recording` or constructs `Recording`; nothing in the
  runner path changed. The only way this breaks anyone is F2.
- Rust's `an_append_failure_does_not_stop_the_conversion` asserts the *joined*
  error rather than only the first, which is a stronger assertion than the test
  name promises and is what kills the Rust join mutant.

## Outside this PR

Clearly separated; none of this should count against the change.

- `append_checkpoint_frames` words the same two failures differently in the two
  languages (`empty cast file` against `<path> is empty: a cast has a header
  line`). Pre-existing; the author names it and routes around it correctly.
- `sanitize_component` returning `"default"` for an empty component where the
  Python `_sanitize` returns `""`. Pre-existing. F3 is only the part of this
  that is newly reachable through a surface this PR adds.
- Python's `UploaderLike` has no way for an uploader to report *why* it declined
  without raising. That asymmetry is the root of F4 and widening it would break
  every existing uploader, so it is correctly not this change.
- Four names for two ideas across the video seams (`VideoBackend` in each
  language, plus `CastVideoConverter` and `VideoConverterLike`). The author
  names it; unifying them is a publisher refactor.
- `RsvgFfmpegBackend.convert` defaults to 24 fps while `VideoConfig.fps`
  defaults to 60. Named in the PR body, and the new tests pin both branches, so
  the discrepancy is at least now covered rather than latent.

## What I could not verify

- I did not run `cargo semver-checks` locally — it is not installed here. F2
  rests on the CI job's own output plus the structural facts I checked directly
  in the source (all-public fields, no `#[non_exhaustive]`, no in-tree literal
  construction), and on the same check being green on the default branch one
  commit earlier.
- I did not exercise the real `rsvg-convert`/`ffmpeg` path end to end; neither
  binary is present here. The "converter binaries missing" case is what I ran
  instead, and both implementations handle it identically.

md-mt:
Thanks — the F1 execution was worth more than the review of the code, and it found the one thing I got wrong by not carrying my own argument one step further. All seven addressed; rebased onto `5d34f1d` (PR 200).

## F1 — step 3 now gets step 1's guard, and I checked 4 and 5 for the same asymmetry

You were right, and the framing "step 1 refuses to trust its closure, step 3 takes its converter on trust" is the whole bug. I had written the reasoning into the step-1 docstring — "the step that lied is the one worth naming" — and then not applied it twice more.

The rule is now general and stated as such, in both docstrings under **"A step is not believed, it is checked"**:

| step | what it claims to produce | how it is checked |
|---|---|---|
| 1, save cast | a cast at the path it was handed | `is_file()` → `save cast: reported success but wrote no file at <path>` |
| 2, append frames | nothing — it edits the cast in place | nothing to check |
| 3, convert | a video at the path it returns | `is_file()` → `convert: reported success but wrote no file at <path>` |
| 4, upload | a URL | an empty string is no URL, treated exactly as a decline |
| 5, record | a `Recording` on a list | cannot fail |

Step 3's failure names step 3 and **skips step 4**, so the upload you observed against a non-existent path no longer happens. Your exact repro, re-run:

```
Python   video: None   url: None   error: 'convert: reported success but wrote no file at .../recording-00-full-session.mp4'   upload calls: []
Rust     video: None   url: None   error: Some("convert: reported success but wrote no file at .../recording-00-full-session.mp4")   uploads: []
```

Step 4 was the other half of the same question and is F5 below. Step 2 genuinely has nothing to check and step 5 appends to a list; both are now stated rather than left implicit.

`silent-encode` is in the corpus beside `silent-save`, as you suggested — these are one guard at three steps and the wording has to match across implementations.

New tests: `a_converter_that_writes_nothing_is_the_convert_step_failing` and its Python mirror, both asserting zero upload calls. Mutation-checked: deleting the guard kills the unit test **and** turns `differential_evidence_manifest` red.

## F2 — filed as breaking; version left alone

Your structural analysis is right and I have not tried to code around it. The changelog entry moved to a new **`### Rust — Changed (breaking)`** heading (ordered after `Rust — Changed`, matching `[0.4.0]`), naming `EvidencePublisher` and the exact symbol, with the constructor call that replaces a struct literal:

```rust
// was
let publisher = EvidencePublisher { dir, identity, renderer, uploader: Some(u) };
// now
let publisher = EvidencePublisher::new(dir, identity).with_renderer(renderer).with_uploader(u);
```

It also records the trade you named: the break is unavoidable given this shape and avoidable with a different one (converter passed to `record_session` rather than hung off the publisher), so the "beside `renderer` and `uploader`" argument is not free.

**No version bump, and `semver-checks` is still red on purpose.** The digit is yours.

## F3 — converged, and now pinned rather than merely unseen

I took the first of your two options and closed it in `record_session` rather than documenting a constraint, because the divergence was in a surface this PR introduces. `_recording_file_stem` applies the same empty-component fallback `store::sanitize_component` does, so `record_session("", ...)` writes `recording-NN-default` on both sides, and an empty-label recording is in the corpus holding it there.

`CapturedStep.file_stem` still differs the same way for the same input — pre-existing, and converging `_sanitize` would move paths that already ship, so it stays a separate change. This narrows the divergence to the surface that predates it instead of extending it to a new one.

## F4 — recorded, not pinned

Correct and I had understated it. `conformance/README.md` now says plainly that `no-url` pins the **shared fallback** and not the shipped behaviour, and that `FallbackUploader` — the only non-test `ArtifactUploader` in the crate — sets `last_error` on every failure, so in practice Rust writes `upload: all uploaders returned no URL` where Python writes `upload: uploader returned no URL`. It sits beside the append note, in the same "deliberately absent, because pinning it would make the harness certify a divergence" list. The probe docstring and the harness module doc say the same.

## F5 — fixed, and it is F1's question at step 4

An empty URL is now no URL, in both implementations: it takes the decline path and records `upload: uploader returned no URL`, with the video kept. I used `is_empty()` / `not url` rather than trimming, deliberately — Rust's `str::trim` and Python's `str.strip()` do not agree on every character, and a whitespace-trimming rule would have been a new divergence in the name of fixing one.

Added `blank-url` to the corpus as a distinct case from `no-url`: both produce the same message, but each implementation rejects the empty string in its own code, so without it the differential harness does not notice one side keeping it. (Verified: it does not — dropping the Rust filter leaves the pair green without this case and turns it red with it.)

## F6 — fixed

`session.save_cast` did not exist. The example copies from `session.cast_path`, which is what a `TerminalSession` actually exposes, and now carries its imports:

```python
collector.record_session(
    "full-session", publisher, lambda dest: shutil.copyfile(session.cast_path, dest)
)
```

You are right about why it drifted — the Rust example is a compile-checked `no_run` doctest and the Python one is prose. Making the Python README examples executable is a repo-wide change, so I have not done it here.

## F7 — qualified

Both docstrings now say where the promise stops: Python catches `Exception`, so `KeyboardInterrupt` and `SystemExit` from any of the three seams propagate and the recording is lost rather than degraded, and a Rust panic unwinds the same way. Phrased as "no failure a step *reports* fails the run", with the boundary stated rather than implied. Behaviour unchanged, as you asked. The same qualification is in the README and the changelog.

## Validation

- **Rust:** 565 tests / 0 failures across 19 binaries; clippy `--all-targets` and `cargo fmt --check` clean.
- **Python:** 849 tests / 0 failures; stdlib-only gate 149 green; `ruff` and `mypy` clean.
- **Corpus** regenerates byte-for-byte, and the Rust half passes against it.
- **Mutation:** the 29 from before all still die, plus 6 new ones for the new behaviour (drop step 3's guard / drop the empty-URL filter / drop the stem fallback, each language). All 35 killed. Of the new Rust three, the differential pair independently kills the step-3 guard and the stem fallback; it does not kill the empty-URL filter, which is why `blank-url` went into the corpus.
- **Rebased** onto `5d34f1d`. The CHANGELOG conflicted with PR 200's entries under the same two headings; resolved by keeping both — `git diff 5d34f1d HEAD -- CHANGELOG.md` shows no deletions.

One honest note on the suites: two of my runs made while `cargo` was compiling in parallel saw the pty-driven `test_demo` time out waiting for its `demo>` prompt. It passes on its own, on this branch and on the pre-change tree, and it touches nothing here. Flagging it rather than quietly re-running until green.

## Unchanged from the "outside this PR" list

`append_checkpoint_frames`' divergent wording, `sanitize_component` itself, `UploaderLike` having no way to report a reason, the four names for two video-seam ideas, and the 24-vs-60 fps default all stay out, as you and I both had them. F3 and F4 are the parts of those that became reachable or checkable here, and only those parts moved.

md-mt:
# PR 196 — re-review after the fix round

**VERDICT:** CHANGES NEEDED — unchanged, because the code is unchanged.
**MERGE RISK:** medium — identical to the first round. The silent false success in step 3 is still there, the public-API break is still filed as an addition, and the branch has since developed a `CHANGELOG.md` merge conflict.
**CI:** 19 pass / 1 fail / 3 skip — the same numbers as the first round. The one failure is still `Public API compatibility (cargo semver-checks)`. I have now **verified that break by compilation** rather than by reading (see F2).
**STALENESS:** **Changed since the first round.** The default branch has moved one commit ahead of this PR's base and `CHANGELOG.md` now conflicts. The conflict is trivial (both sides append under `[Unreleased]`), and I confirmed the merged tree passes everything.

---

## The headline: there is no fix round to review

I was asked to re-review the delta `4ae3ddb..HEAD`. That range is empty.

- The PR head is `4ae3ddbe1e9242a3e051db1dc637a02173a4041c` — the same commit the
  first review was written against.
- `git diff 4ae3ddb HEAD` on the branch produces no output. The tree is
  byte-identical.
- The PR timeline holds four events in total: the two original commits
  (`4e49a39`, `4ae3ddb`), the automated CI report, and the first review comment.
  There is no later commit, no force-push, and no rebase.
- The PR's recorded base is still the commit the branch was cut from, while the
  default branch has moved on by one commit.
- The check run is unchanged, down to the same single failing job.

So **every finding from the first round stands**, not by inference from "no
commits landed" but because I re-ran the decisive ones against this head and got
the same results. Nothing was introduced, and nothing — F1 through F7 — was
addressed or consciously left.

The rest of this report is the work that was still answerable: I re-verified the
headline finding by execution, upgraded the evidence behind the API break from a
CI report to a local compile, answered the two questions the first round did not
(is the asymmetry wider than the step it was reported on, and would the proposed
guard misfire), and re-tested the branch against the moved default branch.

---

## Summary

`record_session` has three caller-supplied seams. **Exactly one of them checks
whether the caller told the truth.** That is the defect, and it is bigger than
"step 3 is missing a check": steps 3 and 4 both take a caller-supplied value and
write it into the manifest as positive evidence without ever confirming it
refers to something real. A fix that patches only step 3 leaves the same shape
one step to its right.

The good news for whoever writes that fix: the guard is safe. I could not
construct a legitimate converter that it would wrongly reject, and the one case
that looks like a false negative — a converter that writes asynchronously — is
**already broken today**, verified.

---

## How I checked

Everything below was run against the PR head in a scratch worktree. Nothing was
pushed and nothing was changed on the branch.

- Confirmed the head SHA, the empty delta, and the four-event timeline through
  the repository's API and through `git`.
- Drove `record_session` through a purpose-built trust probe in **both**
  implementations: a seam per step, each one lying in the specific way that step
  makes possible, reading the resulting `Recording` rather than the code.
- Compiled an external-consumer crate that constructs `EvidencePublisher` by
  struct literal against the default branch and against this PR head.
- Merged the current default branch into the PR head in scratch and ran the full
  gates on the result.

Every finding below says whether I **verified** it by running it or **inferred**
it from reading.

---

## Findings

### F1 (still open) — major — step 3 records a lying converter as a success

**Location:** `python/termproof/collector.py:588-611`;
`rust/crates/termproof/src/evidence/collector.rs:548-570`

**Re-verified by execution at this head, in both implementations.** I built a
converter whose only defect is that it returns the output path without creating
the file — in Rust, through the public `CastVideoConverter::with_runner` seam,
with a runner whose `ffmpeg` exits `0` and writes nothing, which is the shape a
real broken encoder takes — and read the `Recording`:

```
Python
    cast  : recording-00-full-session.cast   exists=True
    video : recording-00-full-session.mp4    exists=False
    url   : 'https://example/v'
    error : None
    upload calls (path, existed_at_upload): [(...mp4, False)]

Rust
    cast  : recording-00-....cast            exists=true
    video : Some("recording-00-....mp4")     exists=Some(false)
    url   : Some("https://example/v")
    error : None
    upload calls (path, existed): [(...mp4, false)]
```

Both control cases with an honest converter produce `exists=True` at the same
positions, so the probe is measuring what it claims to.

`error: null` alongside a `video` path and a `url` is the manifest asserting the
recording succeeded. It did not. This remains the worst outcome the change can
produce, and it is identical in both implementations — so the conformance corpus
cannot see it either.

### The asymmetry is wider than F1 named

**Verified by execution, both implementations.** This is the part of the first
review that needed extending. The defect is not that one step is missing a
check; it is that of the three caller-supplied seams, one is trusted and
verified while two are simply trusted:

| step | caller-supplied seam | the claim it makes | verified? | what happens when the claim is false |
|---|---|---|---|---|
| 1 · save cast | the `save_cast` closure/callable | "I wrote a cast at the path you gave me" | **yes** — the path is checked with `is_file` | named: `save cast: reported success but wrote no file at <path>`; steps 2–4 correctly skipped |
| 2 · append frames | none — the append runs in-tree over the collector's own captured steps | — | n/a, no caller claim exists | — |
| 3 · convert | the publisher's video converter (a public protocol in one implementation, a public runner seam in the other) | "I wrote a video at the path I am returning" | **no** | **silent false success**: `error: null`, a `video` path that does not exist, a `url` for a video never encoded |
| 4 · upload | the publisher's uploader (public in both) | "here is the URL of the file you handed me" | **no** | `url: ''` — or any non-URL string — recorded with `error: null` |
| 5 · attach | none — appends to a list | — | cannot fail | — |

Step 4's hole is F5 from the first round, and it is the same defect one step to
the right, with a smaller blast radius. Verified at this head in both
implementations: an uploader returning `""` yields `url: ""` with `error: null`;
an uploader returning `"not a url at all"` yields exactly that, with
`error: null`.

The two unverified seams are precisely the two whose output reaches the manifest
as **positive** evidence. Step 1 — the one that got the guard — is the only one
whose failure the reader would have noticed anyway, because a missing cast makes
the next two steps fail loudly.

**What I would say to the maintainer:** the decision worth making is not "add
the guard to step 3", it is *which seams in this collector are allowed to be
believed*. Guarding step 3 alone closes the reported hole and leaves the
principle unstated, which is how it will regrow. See also the note in "Outside
this PR" — the same shape already exists in `publish`.

### Question: does the guard introduce a false negative? — verified, no

I built the two candidates and ran them.

**A converter that legitimately writes its output somewhere other than the
requested path** — this is *not* a false negative. The contract is that `convert`
**returns** where the video landed, and the caller reads the returned value, not
the requested one. Verified: a converter that writes `somewhere-else.mp4` and
returns that path produces `video : somewhere-else.mp4 exists=True`, and the
upload is handed a file that exists. A guard applied to the *returned* path
passes this case cleanly. (A guard applied to the *requested* path would break
it — worth stating, since that is the tempting one-liner.)

**A converter that writes asynchronously** — a genuine gap in the guard's model,
but it is **already broken today, without the guard**. Verified: a converter
that returns before the file exists produces an upload call whose `existed`
column is `False`. Step 4 runs synchronously, immediately, on a path with
nothing at it. So an async converter does not work against this API now, and the
guard would not take away anything that currently works — it would convert a
present-day silent breakage into a named failure. That is an improvement, not a
regression.

Both implementations declare the return type as a filesystem path and feed it
straight to `upload(path)`, so "the returned path is a file by the time
`convert` returns" is already the contract in everything but enforcement.

### F2 (still open) — major — the break is real, and I have now proved it locally

**Location:** `rust/crates/termproof/src/evidence/collector.rs:789`;
`CHANGELOG.md:236`

**Upgraded from inferred to verified by compilation.** The first round rested on
the CI job's output plus structural reading. I wrote an external crate that
constructs the publisher by struct literal, exactly as a consumer on the current
release could, and compiled the identical source against both trees:

```
against the default branch:  Finished `dev` profile ... (compiles)

against this PR head:
    error[E0063]: missing field `video_converter` in initializer of `EvidencePublisher`
     --> src/main.rs:7:13
      |
    7 |     let p = EvidencePublisher {
      |             ^^^^^^^^^^^^^^^^^ missing `video_converter`
```

That is the break, reproduced outside CI. It is not an artefact of the checker.

**The changelog is still wrong about it.** Verified at this head: the
`[Unreleased]` section contains `### Rust — Added` and `### Rust — Changed`, and
**no breaking heading at all** (`grep` for "breaking" across `[Unreleased]`
returns nothing). The field addition is filed only under `### Rust — Added`, and
`### Rust — Changed` says, in full, "Nothing." The file's own preamble states
the pre-1.0 rule, and the project has used a `### Rust — Changed (breaking)`
heading before, so the convention exists and is not being followed here.

**The version consequence belongs to the maintainer and I am not deciding it.**
What is not a judgement call is the record: as written, a consumer reading the
released changelog would not learn that their struct literal stopped compiling.
Re-file the entry under a breaking heading naming the exact symbol, and let the
maintainer choose the digit.

### F3–F7 — all still open, none addressed

The tree is byte-identical, so these stand as filed. I re-confirmed the two that
are cheap to check directly at this head:

- **F6 (nit) — verified again.** `python/README.md:130` still calls
  `session.save_cast(dest)`. There is no `save_cast` anywhere in the Python
  package — the only matches are the `record_session` parameter and its
  docstring — and `session` is undefined in the example.
- **F5 (nit)** is verified above, and I would now re-read it as part of the
  step-4 row in the trust table rather than as an isolated nit.
- **F3 (minor)**, **F4 (minor)** and **F7 (nit)** are unchanged code and stand
  exactly as filed in the first round. The filename-stem helpers in both
  implementations are untouched, and the corpus still has no empty-label case
  and no `silent-encode` case beside `silent-save`.

### New — minor — the branch no longer merges cleanly

**Verified.** This is the one thing that genuinely changed since the first
review, and it changed underneath the PR rather than within it.

The default branch has advanced by one commit, and `CHANGELOG.md` conflicts:
both that commit and this PR append entries under `[Unreleased]`. It is an
adjacent-addition conflict, not a semantic one — the resolution is "keep both,
in order".

I resolved it in scratch and ran the gates on the merged tree, so the maintainer
does not have to wonder whether the conflict hides anything:

| gate | merged tree |
|---|---|
| Rust unit + integration + doctests | **562 passed, 0 failed** |
| Python `unittest discover` | **846 passed** |
| standard-library-only gate | **149 passed** |
| `clippy --all-targets` | clean |
| `cargo fmt --check` | clean |
| evidence-manifest corpus regenerated from the oracle | **reproduces byte-for-byte** |
| the differential evidence-manifest harness against the committed corpus | **passes** |

So the rebase is mechanical and safe. It has simply not happened.

---

## What is good

Unchanged from the first round, and I re-confirmed two of these at this head
rather than carrying them forward on trust:

- **The step-1 guard works, and its scope is right.** Verified: a closure that
  returns success and writes nothing is named as `save cast: reported success
  but wrote no file at <path>`, with steps 2–4 correctly skipped and zero upload
  calls. And the guard checking *existence* rather than *validity* is the
  correct depth — I confirmed the deeper check lands where it belongs. With the
  real converter, a closure that writes an empty file produces
  `append checkpoint frames: empty cast file; convert: empty cast file`, and one
  that writes non-cast content produces `convert: expected value at line 1
  column 1`. Bad *content* is caught downstream and named; only a missing
  *file* needed the guard at step 1. Step 3 has the mirror-image gap: it catches
  bad content and misses a missing file.
- **The ordering rule is implemented, not merely documented**, and the
  upload-after-failed-conversion case really is prevented.
- **The conformance pair earns its place** — re-confirmed here: the corpus
  regenerates byte-for-byte from the oracle and the harness passes against the
  committed corpus, including on the merged tree.
- **The blast radius is still nil in-tree.** The only way this reaches anyone is
  the API break in F2.
- **Choosing to treat a lying `save_cast` as step 1 is the subtle call in the
  whole change and it is correct.** F1 exists only because the reasoning stopped
  one step short.

---

## Outside this PR

Clearly separated. None of this counts against the verdict — but the first item
is directly relevant to *how wide* the F1 fix should be, so it is worth the
maintainer's attention when scoping it.

- **`publish` has the identical hole in its renderer seam, and it predates this
  PR.** Verified by execution: a renderer that returns without raising and
  writes no image yields a published step with `screenshot` pointing at a file
  that does not exist, `error: null`, and an upload attempted on the missing
  file. So "believe the caller-supplied seam" is the existing house pattern
  everywhere in this collector, and step 1 of `record_session` is the *new*
  thing — the first seam anyone thought to check. That reframes the fix: the
  question is whether this PR should extend the new principle across
  `record_session`'s own steps (my recommendation, since they are in the diff)
  and file the `publish` case separately, or whether the maintainer wants one
  change that states the principle once for the whole collector.
- The remaining items from the first round are unchanged and still correctly out
  of scope: the two implementations wording the same append failure differently;
  the empty-component sanitiser divergence; the absence of a way for a Python
  uploader to report *why* it declined without raising; four names for two ideas
  across the video seams; and the two written-down default frame rates.

---

## What I could not verify

- I still did not run `cargo semver-checks` locally; it is not installed here.
  That no longer matters for F2 — the break is now reproduced by direct
  compilation of an external consumer against both trees, which is stronger
  evidence than the checker's verdict.
- I did not exercise the real rasteriser/encoder path end to end; neither binary
  is present here. The absent-binaries case and stubbed-runner cases are what I
  ran instead, and both implementations behave identically in them.
- I cannot say *why* the fix round produced no commits — only that it produced
  none. That is for the maintainer to pick up.

---

## Recommendation

Re-dispatch the fix. Concretely, and in priority order:

1. **Give step 3 the guard step 1 has**, applied to the path `convert`
   *returns*: if it is not a file, push `convert: reported success but wrote no
   file at <path>`, leave `video` unset, and skip step 4. Verified safe — no
   legitimate converter is rejected by it.
2. **Decide the principle for step 4 at the same time**, rather than leaving the
   same shape one step over. At minimum, treat an empty URL the way the code
   already treats an empty failure reason.
3. **Re-file the API break under a breaking heading** naming the exact symbol.
   The digit is the maintainer's call.
4. **Add a `silent-encode` case to the corpus beside `silent-save`**, since the
   two guards are then the same guard, and consider an empty-label recording so
   F3's divergence is recorded as known rather than merely unseen.
5. **Rebase.** Mechanical; the merged tree is green on every gate.

md-mt:
# PR 196 — re-review after the fix round

> **Correction to my previous comment on this PR.** That comment was written
> against a stale head: I fetched before the branch was force-pushed, so the
> hash I diffed from was the pre-rebase one and I concluded there was no fix
> round. That was wrong. The fix round exists, it is substantial, and this
> report replaces that comment in full.

**VERDICT:** MERGE WITH NITS
**MERGE RISK:** low — the code is sound and I verified every guard by execution in both implementations. What is left is a version decision, not a defect.
**CI:** 19 pass / 1 fail / 3 skip. The single failure is `Public API compatibility (cargo semver-checks)`, and it is **not** fixable by anything in this PR's diff — see "The one red check" below. The rest is meaningful coverage: both unit suites, the standard-library-only gate, and the differential harness all exercise the new guards, and I confirmed the harness really does kill them (5 mutants, below).
**STALENESS:** Rebased onto the current default-branch tip. No conflicts. All gates green at the head.

---

## Summary

Every one of F1–F7 is addressed, and F1 is addressed the way the reasoning
demanded rather than the way the report happened to word it. The step-3 guard is
real, it is the *same* guard as step 1 — same check, same message — and it is
applied to the path the converter **returns**, which is the choice that avoids
the one false negative worth worrying about. F5 got the same treatment at step
4. F3 was not patched at the reported input; the two implementations were
converged, and I verified agreement across a range of labels, not just the empty
one.

I looked hard for something the fix broke. The one candidate I found — the
shipped fallback uploader producing a divergent step-4 message now that an empty
URL is rejected — does **not** occur; I built the chain and ran it.

The nits are the PR description, which still describes the pre-fix change, and
the version consequence of the API break, which is the maintainer's to make and
is what keeps the board amber.

---

## How I checked

Against the current head in a scratch worktree. Nothing pushed, nothing changed
on the branch. Every claim below says whether I **verified** it by running it or
**inferred** it from reading.

- Drove `record_session` through a trust probe in **both** implementations, one
  lying seam per step, reading the resulting `Recording` rather than the code.
- Mutation-tested all five new guards: reverted each in turn, recorded which
  test went red and whether the conformance pair also went red.
- Compiled an external consumer crate that builds `EvidencePublisher` by struct
  literal, against the default branch and against this head.
- Ran the full gates at the head, and regenerated the conformance corpus from
  the oracle.

---

## Findings

### F1 — **fixed**, and it is the same guard — verified by execution, both implementations

I made the converter lie — return the output path having created nothing; in
Rust through the public `with_runner` seam, with a runner whose `ffmpeg` exits
`0` and writes nothing, which is what a real misconfigured encoder looks like —
and read the `Recording`:

```
Python                                   Rust
  video : None                             video : None
  url   : None                             url   : None
  error : 'convert: reported success       error : "convert: reported success
           but wrote no file at <path>'             but wrote no file at <path>"
  uploads: []                              uploads: []
```

Identical wording to step 1's message, `video` left unset, step 4 skipped, zero
upload calls. Honest-converter controls at the same positions produce a real
file, a URL and `error: None`, so the probe measures what it claims to.

### The asymmetry, step by step — verified

The first round's real point was that the *sequence* trusted some callers and
not others. It no longer does. Of the three caller-supplied seams, all three that
can be checked now are:

| step | caller-supplied seam | the claim | checked now? |
|---|---|---|---|
| 1 · save cast | the closure/callable | "I wrote a cast at the path you gave me" | **yes** — `is_file` on the path, since the first commit |
| 2 · append frames | none — runs in-tree over the collector's own captured steps | — | n/a; edits the cast in place and produces nothing to check |
| 3 · convert | the publisher's converter | "I wrote a video at the path I am returning" | **yes, new** — `is_file` on the **returned** path |
| 4 · upload | the publisher's uploader | "here is a URL for it" | **yes, new** — an empty string is treated as no URL |
| 5 · attach | none | — | cannot fail; appends to a list |

Step 4 verified in both: an uploader returning `""` now yields
`upload: uploader returned no URL` with `url` unset and the **video kept** —
which is right, since step 4's failure costs the URL and nothing else. A
non-empty non-URL string is still accepted, which is correct: there is nothing
local to check it against, and inventing a URL validator here would be worse
than the problem.

The principle is now also written down — "A step is not believed, it is checked"
appears in both docstrings, both READMEs and the changelog — which is what makes
this a lesson learned rather than a patch applied.

### Did the guard introduce a false negative? — verified, no

Two candidates, both built and run:

**A converter that legitimately writes elsewhere and returns that path.** Not
rejected. Verified in Python (the protocol allows a converter to choose its own
path): `video : somewhere-else.mp4 exists=True`, uploaded successfully,
`error: None`. The guard checks the **returned** path, so a truthful return is
accepted wherever it points. This is the detail that matters — a guard on the
*requested* path would have broken this case, and it is the tempting one-liner.

**A converter that writes asynchronously.** Now recorded as
`convert: reported success but wrote no file`. This is the one behaviour change
in the fix, and it is worth stating plainly: an async converter was **already
broken** before this PR's fix commit — I verified in the previous round that
step 4 was handed a path with nothing at it (`existed_at_upload = False`). The
guard turns a silent breakage into a named one. Nothing that worked stops
working.

### F5 — **fixed** — verified, both implementations

Covered in the table above. I also checked the thing most likely to have gone
wrong: Rust routes a rejected empty URL into the `last_error()` branch, and
Python has no `last_error`, so a Rust uploader that returns `""` *and* reports a
`last_error` would write a different message than Python. **Verified that the
shipped uploader does not do this**: I built a fallback chain of
`[fails-with-a-reason, returns-""]` and ran it —

```
    url   : None
    error : Some("upload: uploader returned no URL")   // identical to Python
```

because `FallbackUploader::upload` clears `last_error` at the top of every call
and the degraded-success path does not set it. The divergence is only reachable
by a third-party uploader that returns `Some("")` while also reporting a
`last_error`, which contradicts that method's own documented contract, and it
falls inside the caveat the author now records (see F4). **No new divergence.**

### F3 — **fixed, and converged rather than patched** — verified

Python now applies the same empty-component fallback Rust's sanitiser applies. I
compared the two implementations' stems across a range of inputs rather than
only the reported one:

| label | Python | Rust |
|---|---|---|
| `""` | `recording-00-default` | `recording-00-default` |
| `"!!"` | `recording-00---` | `recording-00---` |
| `"full-session"` | `recording-00-full-session` | `recording-00-full-session` |
| `"   "` | `recording-00----` | `recording-00----` |
| `"a b"` | `recording-00-a-b` | `recording-00-a-b` |
| `"日本"` | `recording-00-日本` | `recording-00-日本` |

Exact agreement on all six. An empty-label recording is now in the corpus, so
the pair holds it there. The fallback is applied in `_recording_file_stem`
rather than in the shared sanitiser, with the reasoning written down — that is
the right call: converging the sanitiser itself would move paths that already
ship, and this narrows the divergence to the pre-existing surface instead of
extending it to a new one.

### F4 — **addressed as documentation**, and the documentation is accurate — verified

`conformance/README.md` and the probe docstring now say that `no-url` and
`blank-url` pin the *shared fallback* and not the shipped behaviour, naming the
concrete consequence. I checked the claim rather than the prose: a
`FallbackUploader` whose only uploader fails with a reason produces
`upload: primary store refused`, where Python for the same shape produces
`upload: uploader returned no URL`. Real, reachable, and correctly recorded in
prose rather than frozen into the corpus.

### F2 — **filed correctly**; the break itself re-verified by compilation

The changelog now carries `### Rust — Changed (breaking)` under `[Unreleased]`,
naming the exact symbol, the exact lint (`constructible_struct_adds_field`), the
migration (`EvidencePublisher::new(...).with_renderer(...).with_uploader(...)`),
and the shape trade-off that made the break unavoidable *given this design*.
That is a better entry than I asked for.

The break is real, and I proved it locally rather than trusting the CI job — the
same external consumer source compiles against the default branch and fails
against this head:

```
error[E0063]: missing field `video_converter` in initializer of `EvidencePublisher`
```

I also checked the fix round added no *further* public-API surface: no
`pub fn`/`struct`/`trait`/`const` was added or changed anywhere in the Rust
delta. The single red check is still the one field.

### F6 — **fixed** — verified

The Python README example now reads
`lambda dest: shutil.copyfile(session.cast_path, dest)`, with the imports it
needs and a comment saying what `session` is. `session.cast_path` is a real
attribute; the old `session.save_cast(...)` was not.

### F7 — **fixed, and the claim is true** — verified

The boundary is now stated in both docstrings, both READMEs and the changelog.
The docstrings make a factual assertion, so I ran it rather than reading it:

- Rust: a panic in the closure escapes `record_session` (`true`) and the
  collector ends with `0` recordings attached.
- Python: a `KeyboardInterrupt` from the closure escapes (`true`) and the
  collector ends with `0` recordings.

The line is drawn in the same place in both, exactly as documented.

### Mutation check — I killed all five new guards, and the pair killed four

Not taken on the PR's word; applied and reverted one at a time.

| mutant | test that went red | conformance pair |
|---|---|---|
| delete the Rust step-3 guard | `a_converter_that_writes_nothing_is_the_convert_step_failing` | **also red** — the `silent-encode` case flipped to `error: null` with a video and URL |
| delete Rust's empty-URL filter | `an_empty_url_is_no_url` | **also red** |
| delete the Python step-3 guard | `test_a_converter_that_writes_nothing_is_the_convert_step_failing` | **corpus changed** (would turn the Rust half red) |
| Python empty URL accepted again | 1 failure in `test_collector` | **corpus changed** |
| delete Python's empty-label fallback | `test_an_empty_label_names_a_recording_the_way_rust_names_it` | **corpus changed** |

Each killed by the test that should own it. The new corpus cases are load-bearing,
not decorative.

### The rebase — clean, and the merged state is green — verified

The branch now sits directly on the current default-branch tip. `git merge-base`
confirms it, there is nothing to merge back down, and `git merge-tree` reports
no conflict. The `CHANGELOG.md` conflict that had opened up against the moved
default branch is gone, resolved in the rebase.

Gates at the head:

| gate | result |
|---|---|
| Rust unit + integration + doctests | **565 passed, 0 failed** |
| Python `unittest discover` | **849 passed** |
| standard-library-only gate | **149 passed** |
| `clippy --all-targets` | clean |
| `cargo fmt --check` | clean |
| evidence-manifest corpus regenerated from the oracle | **reproduces byte-for-byte** |
| differential evidence-manifest harness | **passes** |

---

## The one red check, and what it actually needs

Worth being precise, because it is the only thing standing between this PR and a
green board, and it is **not** a changelog problem — F2 fixed the record, and
the record is now right.

The job runs `cargo semver-checks check-release -p termproof`, which compares
the working tree against **the version published on the registry**. That
published version is `0.4.0`, and the workspace still declares `0.4.0`. So the
checker is comparing a tree that adds a public field against an already-released
crate at the same number, and it will stay red until the declared version moves.
The job's own comment says this is by design: it "fails on any breaking change
to a public API — under 0.x that is any change the pre-1.0 convention would call
a minor bump."

The changelog preamble states that convention: under `0.x`, a breaking change
bumps the minor digit. So the check is doing exactly its job, and the only thing
that clears it is bumping to `0.5.0`.

**That decision is the maintainer's and I am not making it.** The options I can
see are: bump the workspace version in this PR and turn the board green; or
leave it under `[Unreleased]` and bump at release time, accepting a red advisory
check on the PR. Both are defensible. What is not tenable is the current PR
description, which still says "No version bump" with no mention of the
classification — see the nit below.

---

## Nits

- **nit — the PR description still describes the pre-fix change.** It has no
  mention of the fix round, still says "No version bump" with no reference to
  the now-filed breaking change, and its Validation block quotes 549 Rust / 841
  Python tests where the head has 565 / 849. Nothing in the code depends on
  this, but the description is what a reader reaches for first, and right now it
  contradicts the changelog on the one point that needs a decision.

- **nit — worth naming in the docstring, since the neighbouring boundary now
  is.** The step-3 guard checks that the returned path *is a file*, not that it
  is a video. That is the right depth — a truncated or corrupt encode is not
  something this method can adjudicate — and the sibling case is already handled
  well (a cast that is not a cast is caught downstream and named: I verified a
  real converter over an empty cast produces
  `append checkpoint frames: empty cast file; convert: empty cast file`, and
  over garbage, `convert: expected value at line 1 column 1`). One sentence
  saying the check is existence, not validity, would close the last gap between
  what the docstring promises and what the guard does.

---

## What is good

- **The fix generalised instead of patching the report.** F1 named step 3; the
  change fixed step 3 *and* step 4, converged F3 rather than special-casing it,
  and wrote the principle down in five places. That is the difference between
  closing a finding and learning from it.
- **The guard is on the returned path**, which is the one detail that decides
  whether it is correct or a new bug. Verified.
- **The new corpus cases are load-bearing.** `silent-encode`, `blank-url` and
  the empty-label recording each kill a mutant on their own.
- **F4 was handled by recording the disagreement rather than pinning it.**
  Freezing it into the corpus would have made the harness certify a divergence;
  the prose says what the corpus does and does not cover, and I confirmed the
  prose is accurate against the shipped uploader.
- **The changelog entry for the break is better than the finding asked for** —
  lint name, migration snippet, and an honest account of the shape trade-off.
- **The boundary docstrings are true.** I ran both.

---

## Outside this PR

Clearly separated; none of it counts against the verdict.

- **`publish`'s renderer seam still takes its caller on trust.** Verified by
  execution: a renderer that returns without raising and writes no image yields
  a published step with `screenshot` pointing at a file that does not exist,
  `error: null`, and an upload attempted on the missing file — the exact shape
  F1 described, one method over. It is pre-existing and outside this diff, but
  it is now the *only* seam in this collector that is still believed, which
  makes it more conspicuous than it was. Worth its own issue.
- The remaining pre-existing items are unchanged and correctly out of scope: the
  two implementations wording the same append failure differently;
  `CapturedStep.file_stem` still diverging for an empty component (this PR
  narrows the divergence rather than extending it, and says so); four names for
  two ideas across the video seams; and the two written-down default frame
  rates.

---

## What I could not verify

- I did not run `cargo semver-checks` locally; it is not installed here. It no
  longer matters — the break is reproduced by compiling an external consumer
  against both trees, and the version arithmetic above is checked against the
  registry and the declared workspace version.
- I did not exercise the real rasteriser/encoder path end to end; neither binary
  is present here. Stubbed runners and the absent-binaries case are what I ran
  instead, and both implementations behave identically in them.

md-mt:
## The semver break is waived, deliberately — 0.4.x stands

Per the captain's decision on this PR: **termproof stays on 0.4.x and the `EvidencePublisher` break is waived, not fixed.** No version bump, no change to the struct, and the `Public API compatibility (cargo semver-checks)` job stays enabled and stays gating.

### What the waiver is

One lint, via cargo-semver-checks' own mechanism, in `rust/crates/termproof/Cargo.toml`:

```toml
[package.metadata.cargo-semver-checks.lints]
constructible_struct_adds_field = "allow"
```

The comment above it names the break, the decision, and the release line it was granted for.

Measured, not asserted — `cargo semver-checks check-release -p termproof` 0.50.0, the same version CI pins:

| | checks run | result |
|---|---|---|
| before | 196 | `1 fail` — `constructible_struct_adds_field`, `field EvidencePublisher.video_converter` |
| after | 195 | `no semver update required` |

So 195 of 196 checks still gate this crate. One lint is no longer asked; nothing else was touched.

### Why not the other options

- **Not a version bump** — that is the decision, and it is not mine.
- **Not `#[non_exhaustive]`** — that would be fixing it, and it is itself a breaking change.
- **Not disabling the job or the workflow path** — that would waive every future break, including ones nobody has looked at.

### The CHANGELOG entry still describes the break

Unchanged, apart from a paragraph recording the decision. **A waiver decides what the version does about a break, not whether the break happened** — a consumer whose `EvidencePublisher { .. }` literal stopped compiling needs that entry and the migration snippet whatever the digit says. It stays under `### Rust — Changed (breaking)`.

### Two guards, because the mechanism is coarser than the decision

cargo-semver-checks has no per-item allow, so a lint-level waiver would also cover a new `pub` field on *any* exhaustively-constructible struct in the crate. That is broader than what was decided, so the difference is bounded in-tree:

- **`every_public_field_of_the_publisher_is_accounted_for`** builds `EvidencePublisher` from an exhaustive struct literal. A sixth field stops it compiling. That recovers, for the struct the waiver was granted for, exactly the cover the lint gave it.
- **`the_semver_waiver_is_scoped_to_the_release_it_was_granted_for`** fails the build if the version leaves `0.4.x` while the waiver is still present — the waiver was granted on the reasoning that 0.4.x stands, so it expires when that does, and has to be re-decided rather than carried along. (It matches on the lint name alone: I checked the packaged tarball, and `cargo package` strips the comments but keeps the key, so the formatting is not mine to rely on.)

The Cargo.toml comment says to delete the waiver and both tests together, or neither.

### Verified by execution

| | result |
|---|---|
| failure reproduces before the waiver, naming the field | ✅ |
| waiver clears it, 195 checks still run | ✅ |
| **removing the waiver brings the failure back** — so the waiver is what suppresses it, not something else | ✅ |
| a sixth public field on `EvidencePublisher` fails to compile | ✅ |
| version set to `0.5.0` with the waiver present fails the scope test | ✅ |
| `cargo package` + `verify-package-contents.sh` still pass; the waiver key survives manifest normalisation | ✅ |

Rust 567 tests / 0 failures, clippy and fmt clean. Python 849 / 0, `ruff` and `mypy` clean — no Python code changed in this commit.

### Also

`rust/docs/publishing.md`'s release checklist now says to read the waiver list *before* trusting the green check, with the reason: a passing run means nothing broke **except what a waiver already covers**. Today there is exactly one entry, and it names #196 and 0.4.x.

Rebased check: `origin/main` has not moved since the last push, so the branch is current at `3483333`.

PR #200, opened 2026-08-20
result: one scoring rule, and the empty-assertion case as contract (#198)

### LINKED ISSUE #198: Scoring a bool assertion map is generic, but every consumer writes it
`RunResult` carries `score: f64` alongside `assertions`, so "what fraction of
assertions held" is already the crate's notion of a score. There is no function
that computes it, so consumers write their own.

Verbatim from one consumer:

    /// Fraction of truthy assertion values (1.0 when empty).
    pub fn score_from(assertions: &BTreeMap<String, bool>) -> f64 {
        if assertions.is_empty() { return 1.0; }
        let passed = assertions.values().filter(|v| **v).count();
        passed as f64 / assertions.len() as f64
    }

...plus a small companion that builds the map from `(name, bool)` pairs.

Nothing here is product-specific, and the empty-set case is exactly the kind of
decision worth making once in a library rather than independently in every
caller. Scoring `1.0` for a run that asserted nothing is defensible — but a
consumer that chose `0.0` instead would produce results that look comparable to
ours and are not.

**Suggestion.** `result::score_from(&BTreeMap<String, bool>) -> f64` (and the
Python equivalent), with the empty-set behaviour documented as part of the
result contract rather than left to each caller.

md-mt:
Closes #198.

## What changed

`RunResult.score` is already the crate's notion of "what fraction of assertions held", but
nothing computed it from a name → passed map, so every consumer wrote its own four lines.
The arithmetic was never the problem — **the empty case is a contract decision** that each
caller was making independently. A run that asserted nothing scoring `1.0` is defensible; a
consumer that chose `0.0` would publish numbers that look comparable to these and are not.

Added, with identical semantics in both languages:

| | Rust | Python |
|---|---|---|
| score a map | `result::score_from(&BTreeMap<String, bool>) -> f64` | `models.score_from(Mapping[str, bool]) -> float` |
| build the map | `result::assertion_map(pairs)` | `models.assertion_map(pairs)` |

Both agree on: empty → `1.0`; ordering cannot move the result; a repeated name is one entry
whose **last** value wins. The difference from the existing list-shaped
`score_from_assertions` is spelled out in the docs — a list weighs duplicate names once
each, a map folds them into one.

## The empty case is documented in the result contract

Not only on the new function:

- **Rust** — the `RunResult::score` field doc, i.e. on the field the number ends up in.
- **Python** — the `RunResult` class docstring (the file has no module docstring; this is
  the nearest equivalent to the Rust field doc).
- **Spec** — `spec/003-builtin-assertions/spec.md` FR-022, which is where the two
  implementations' shared contract for `score` already lived. It now says why `1.0` and
  names every function that implements the rule.

## One path, not a second one

The crate *did* compute scores inline — in three places once this change was counted:

- `RunResult::score_from_assertions` (Rust)
- `assertions::score` (Rust) — a separate hand-written copy of the same arithmetic
- `models.score_from_assertions` (Python)

All of them now call one private rule (`result::score_of` / `models._score_of`). Behaviour
is unchanged: the assertion differential harness reproduces the committed corpus, `score`
cases included. `assertions::score` keeps its signature over the legacy
`models::AssertionResult`; only its body moved.

## Tests

New tests in both languages cover empty, all-true, all-false, mixed, ordering independence,
duplicate names, and list/map agreement. Each was broken to confirm it goes red:

| mutation | red in Rust | red in Python |
|---|---|---|
| empty → `0.0` | `an_empty_assertion_map_scores_one_not_zero` | `test_an_empty_assertion_map_scores_one_not_zero` |
| fraction → always `1.0` | `a_map_scores_the_fraction_that_held` | `test_a_map_scores_the_fraction_that_held` |
| `assertion_map` keeps the first value | `a_repeated_name_is_one_assertion_and_the_last_value_wins` | `test_a_repeated_name_is_one_assertion_and_the_last_value_wins` |
| list path stops delegating (empty list → `0.0`) | `the_list_and_the_map_score_the_same_assertions_alike` | `test_the_list_and_the_map_score_the_same_assertions_alike` |
| score counts only the first two values | `a_map_scores_the_fraction_that_held` | `test_the_order_...`, `test_a_map_scores_...` |

**One test stayed green where its Python mirror went red**, and it is worth naming: the Rust
`the_order_the_pairs_arrived_in_cannot_move_the_score` cannot be made to fail by any mutation
of the scoring rule, because `BTreeMap` iterates in key order whatever order the pairs
arrived in — the guarantee holds by construction, and the test states it rather than guards
it. Python's `dict` preserves insertion order, so the same test there is a real guard and does
go red under the truncation mutant.

Also run: full `cargo test --all-features` (all green, differential harnesses included),
`cargo clippy --all-targets --all-features`, `cargo fmt --check`, the full 830-test Python
suite, `ruff check`, `mypy termproof/models.py`.

## `cargo semver-checks`

Clean — `196 checks: 196 pass, 58 skip`, "no semver update required". The change is purely
additive; no version number was touched.

## Deliberately not done

- **No new conformance corpus.** The three differential harnesses compare steps, assertions
  and the evidence manifest. Scoring is already covered by the assertions pair — six
  `003-FR-022` `score/*` cases whose expectations `differential_assertions.rs` checks through
  `assertions::score`, which this change routes into the shared rule — so parity for the rule
  is asserted where it already was. The map-shaped surface has no I/O and no corpus of its
  own; its parity is asserted by the mirrored unit tests above. The evidence-manifest pair is
  untouched because nothing about evidence changed.
- **Not exported from `termproof/__init__.py`.** `score_from_assertions` is not either;
  consumers import from `termproof.models`, which mirrors `termproof::result` being reachable
  as a module rather than re-exported at the crate root. Adding both to the package root
  would be a reasonable follow-up, in both implementations at once.

## Noticed, not fixed

- `models::RunResult` (Rust) is the legacy twin of `result::RunResult` and duplicates every
  field. Its `score` doc now points at the canonical contract rather than restating it, which
  is the smallest thing that stops the two docs drifting; collapsing the twins is its own
  change.
- The `[Unreleased]` → `Rust — Changed` section opened with a bullet reading "Nothing." That
  became false with this PR, so it now reads "On SVG geometry, nothing." — content unchanged
  otherwise. Flagging it because it edits a sibling PR's entry.
- `ruff format --check` fails on `python/termproof/models.py` and `python/tests/test_models.py`
  on **pre-existing** lines (it wants to join wrapped expressions that predate this change).
  Not touched; `ruff check` passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,20 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
 
 ### Python — Added
 
+- **`termproof.models.score_from` and `termproof.models.assertion_map`:**
+  `score_from` scores a name → passed mapping the way `RunResult.score` is
+  defined — the fraction that held — and `assertion_map` builds that mapping
+  from `(name, passed)` pairs, last pair for a name winning. Consumers were
+  each writing those four lines, which was fine for the arithmetic and not fine
+  for the one decision inside it: **a run that asserted nothing scores `1.0`,
+  not `0.0`**, and a consumer that chose the other way published numbers that
+  looked comparable to these and were not. That choice is now stated in the
+  `RunResult` docstring and in `spec/003-builtin-assertions/spec.md` FR-022,
+  as part of the result contract rather than as a note on one function.
+  `score_from_assertions` is unchanged in behaviour and now delegates to the
+  same rule; the difference between the two shapes is that a list weighs
+  duplicate names once each and a mapping folds them into one entry.
+
 - **`termproof.cast_video.append_checkpoint_frames`:** appends the captured
   checkpoint screens to a cast as held trailing frames, so a recording ends by
   replaying the evidence sequence instead of stopping on whatever the last
@@ -197,6 +211,12 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
 
 ### Rust — Added
 
+- **`result::score_from` and `result::assertion_map`:** the same two functions
+  with the same semantics, `score_from` taking `&BTreeMap<String, bool>` and
+  `assertion_map` collecting `(impl Into<String>, bool)` pairs. The empty-map
+  case is `1.0` here too, and is documented on `RunResult::score` — the field
+  the number ends up in — rather than only on the function that computes it.
+
 - **`evidence::cast_video::append_checkpoint_frames`:** the same capability with
   the same semantics, taking `hold_seconds: Option<f64>` where Python takes a
   defaulted keyword and returning `Err` where Python raises `ValueError`. The
@@ -209,7 +229,14 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
 
 ### Rust — Changed
 
-- Nothing. `SvgMetrics` in `terminal::attributed` already took every default
+- **scoring:** `assertions::score` and `RunResult::score_from_assertions` were
+  two hand-written copies of the same arithmetic, and `result::score_from`
+  would have made a third. All three now call one private rule in `result`.
+  Nothing observable moved — the assertion differential harness reproduces the
+  committed corpus, `score` cases included — and the point is that the
+  empty-set answer can no longer drift between them.
+
+- On SVG geometry, nothing. `SvgMetrics` in `terminal::attributed` already took every default
   from the `DEFAULT_*` constants beside it, `ScreenshotRenderer` and
   `CastVideoConverter` already referenced those same constants rather than
   restating them, and the Rust stack was already the Linux-first one — the
--- a/python/termproof/models.py
+++ b/python/termproof/models.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections.abc import Iterable, Mapping
 from dataclasses import dataclass, field, replace
 from pathlib import Path
 from typing import Any
@@ -152,6 +153,22 @@ class PublishedArtifact:
 
 @dataclass(frozen=True)
 class RunResult:
+    """One recipe × one renderer, as a canonical result payload.
+
+    ``score`` is the fraction of assertions that held, in ``0.0..=1.0``.
+    **A run that asserted nothing scores 1.0, not 0.0.** That is part of this
+    contract rather than a choice each producer makes, because a score is read
+    comparatively: a consumer that scored the empty set ``0.0`` would publish
+    numbers that look like these and rank against them wrongly. The reading is
+    "no assertion failed", which is what an empty set reports; "nothing was
+    verified" is carried by ``assertions`` being empty, and a caller that cares
+    about it should read that rather than the score.
+
+    :func:`score_from` computes it from a name → passed mapping and
+    :func:`score_from_assertions` from an :class:`AssertionResult` list; both
+    are the same rule, and neither should be re-derived by a caller.
+    """
+
     recipe_name: str
     passed: bool
     exit_code: int | None
@@ -243,11 +260,44 @@ def load_recipe(path: Path) -> Recipe:
     return replace(recipe, source_path=str(path))
 
 
+def score_from(assertions: Mapping[str, bool]) -> float:
+    """:attr:`RunResult.score` for a name → passed mapping.
+
+    **An empty mapping scores 1.0.** See :class:`RunResult` for why that is the
+    contract and not each caller's decision.
+
+    The mapping's ordering cannot affect the result — only how many values are
+    true and how many there are — so a caller that collected its assertions in
+    a different order gets the same number. Two entries under one name are one
+    entry, which is the difference from :func:`score_from_assertions`: a list
+    keeps duplicates and weighs each of them.
+    """
+    return _score_of(assertions.values())
+
+
+def assertion_map(pairs: Iterable[tuple[str, bool]]) -> dict[str, bool]:
+    """The mapping :func:`score_from` scores, from ``(name, passed)`` pairs.
+
+    The last pair for a name wins, as assigning into the mapping would.
+    """
+    return {str(name): bool(passed) for name, passed in pairs}
+
+
 def score_from_assertions(assertions: list[AssertionResult]) -> float:
-    if not assertions:
+    """:attr:`RunResult.score` from assertion results (all pass → 1.0, else fraction)."""
+    return _score_of(assertion.passed for assertion in assertions)
+
+
+def _score_of(values: Iterable[bool]) -> float:
+    """The one scoring rule, over whatever the caller counted as an assertion."""
+    total = 0
+    passed = 0
+    for value in values:
+        total += 1
+        passed += 1 if value else 0
+    if total == 0 or passed == total:
         return 1.0
-    passed = sum(1 for assertion in assertions if assertion.passed)
-    return 1.0 if passed == len(assertions) else passed / len(assertions)
+    return passed / total
 
 
 def _normalize_renderers(value: Any) -> dict[str, list[str]]:
--- a/python/tests/test_models.py
+++ b/python/tests/test_models.py
@@ -2,7 +2,14 @@ from __future__ import annotations
 
 import unittest
 
-from termproof.models import AssertionResult, RunResult, StepResult
+from termproof.models import (
+    AssertionResult,
+    RunResult,
+    StepResult,
+    assertion_map,
+    score_from,
+    score_from_assertions,
+)
 
 
 def _result() -> RunResult:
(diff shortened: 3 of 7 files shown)


md-mt:
# PR 200 — result: one scoring rule, and the empty-assertion case as contract (#198)

**VERDICT:** MERGE
**MERGE RISK:** low — purely additive surface over a rule whose behaviour I confirmed
unchanged in both languages; the only way this bites is a reader who meets `score` at
`parity::RecipeOutcome` and still sees no empty-set rule.
**CI:** 20 pass / 3 skip. Meaningful for this change, and stronger than the PR body claims —
see "What is good". I additionally reproduced the full Rust suite (`--all-features`, all
green), the 830-test Python suite (OK), clippy, `cargo fmt --check`, the stdlib-only Python
job (145 tests OK, 6 skipped) and `cargo semver-checks` locally.
**STALENESS:** Current. The default-branch tip is an ancestor of the PR head
(`git merge-base --is-ancestor` succeeds), so it applies cleanly and I tested against the
real merge state.

## Summary

Two additive functions in each language plus one private rule that the three previously
hand-written scoring copies now call. I ran both implementations rather than reading them:
they are **bit-identical** across 47,346 cases. The four load-bearing tests each go red
under a targeted mutation. The empty-set decision genuinely landed in the result contract,
in three places, not on the new function's docstring. I found no surviving second scoring
path. The findings below are all nit-to-minor.

## Findings

### 1. minor — the one remaining place the crate documents `score` without the contract

**Location:** `rust/crates/termproof/src/parity.rs:67`

```rust
/// Score in `0.0..=1.0`.
pub score: f64,
```

`RecipeOutcome::score` is populated from `RunResult::score` (`parity.rs:194` lists `score`
among the fields the summary reads), so this is the same number under a different type.
The PR updated the exactly-analogous bare-range doc on the legacy twin —
`models.rs:176`, `/// Score 0..1.` → a pointer at `result::RunResult::score` — which makes
leaving this one the inconsistency rather than the default.

**Failure case:** a consumer writing a comparison against `RecipeOutcome` reads
`0.0..=1.0`, sees nothing about the empty set, and re-derives the decision the PR exists to
centralise — which is the exact scenario issue #198 describes. Low likelihood (the value is
transparently copied, not produced), which is why this is minor and not major.

**What I would do:** one line, the same pointer `models.rs:176` got. Not worth blocking.

**Verified** — read the file, confirmed no Python counterpart module exists, and confirmed
by grep that this is the *only* remaining `score` doc in either implementation that states
a range without the rule.

### 2. nit — the Rust ordering test is a statement, not a guard

**Location:** `rust/crates/termproof/src/result.rs`,
`the_order_the_pairs_arrived_in_cannot_move_the_score`

The author discloses this in the PR body and the disclosure is **exactly right** — I
checked rather than took it. Under a mutation that makes the rule order-sensitive (count
only the first two values), the results were:

| | Rust | Python |
|---|---|---|
| `..._order_the_pairs_arrived_in_cannot_move_the_score` | **green** | **red** |
| `a_map_scores_the_fraction_that_held` | red | red |

`BTreeMap` normalises to key order, so `forwards` and `backwards` are the *same map* by the
time `score_from` sees them, and the assertion cannot fail for any implementation of the
rule. That is the "asserts something that could never fail" shape — mitigated entirely by
the fact that the author named it in the PR body instead of letting the next person
discover it. Leaving it is defensible; it documents a guarantee that holds by construction.

**Verified** by mutation, both languages.

### 3. nit — the two new Rust surfaces sit at different altitudes from each other

`result::score_from` is a free function; `RunResult::score_from_assertions` is an
associated function on the type. Both compute the same contract from different shapes. In
Python both are module-level, so the shapes agree there and disagree in Rust. The issue
asked for `result::score_from` by name, so this matches the request — flagging only because
a consumer who knows `RunResult::score_from_assertions` will look for
`RunResult::score_from` and not find it. Taste; I would not change it now.

**Verified** by reading the diff.

### 4. nit — changelog places the same internal change in different sections per language

The Rust internal unification gets its own `Rust — Changed` bullet. The identical Python
unification (`score_from_assertions` now delegating) is folded into the `Python — Added`
bullet instead, even though a `Python — Changed` section exists in `[Unreleased]`. Both
readings are defensible — nothing observable moved, so "Added" alone is honest — but the
project keeps mirrored language sections and this breaks the mirror.

On the substantive question: **the changelog files this correctly.** It is a pure addition,
filed under Added in both languages, and I reproduced `cargo semver-checks` myself:
`196 checks: 196 pass, 58 skip`, "no semver update required". No break is being filed as an
addition here.

**Verified** — ran semver-checks; read both `[Unreleased]` sections.

### 5. nit — `assertion_map` coerces in Python where Rust cannot

Python does `{str(name): bool(passed) for ...}`; Rust takes `S: Into<String>` and a real
`bool`. So `assertion_map([("a", 1)])` is accepted in Python and would not compile in Rust.
Only reachable by violating the declared `Iterable[tuple[str, bool]]`, so it is not a parity
break in practice — noting it because the pair is meant to be semantically identical.

**Verified** by reading; not exercised, since reaching it requires ignoring the annotation.

## What is good

**The empty-set decision is genuinely in the contract, in three places.** I checked where it
landed rather than trusting the PR body:

- `result.rs` — on the `RunResult::score` **field**, i.e. on the thing the number ends up
  in, not on the function that computes it.
- `models.py` — the `RunResult` class docstring (the file has no module docstring; this is
  the nearest equivalent).
- `spec/003-builtin-assertions/spec.md` FR-022 — the shared cross-implementation contract,
  which now says *why* `1.0` and names every function implementing the rule.
- Plus the legacy Rust twin at `models.rs:176`, redirected rather than restated, which is
  the right call for two docs that would otherwise drift.

Finding 1 is the sole gap, and it is a derived copy.

**The two implementations agree exactly — I ran them, and the answer is bit-identical.** I
built a sweep harness emitting raw IEEE-754 bits from both and diffed:

- every `(passed, total)` for `total` in `0..=300` — 45,451 cases, covering empty, all-true,
  all-false and every mixed fraction
- the same for `total` in `0..=60` with the pairs inserted in reverse order — ordering
  independence, 1,891 cases
- all four duplicate-name combinations

`47,346` lines, `diff` reports **zero** differences. Not "agrees to within an epsilon" —
the same 64 bits, including the irrational-in-binary cases (`1/3` → `3fd5555555555555`,
`1/7` → `3fc2492492492492`) where a last-place divergence would show up first. Both sides
divide two exactly-representable integers as doubles, so this is structural rather than
lucky, but it is now checked.

**No second scoring path survives.** I searched myself rather than taking the claim. On the
default branch there were the three copies the PR names; on the PR head, the entire
repository contains exactly one fraction computation per language
(`result.rs:229`, `models.py:300`). The two other `filter(|r| r.passed).count()` sites I
turned up (`evidence/report.rs:61,123`) are run-level "X/Y passed" counts for a Markdown
header and a CLI summary line — not scoring, correctly left alone.

**The tests are real.** I broke the code four ways and confirmed each goes red in both
languages:

| mutation | Rust | Python |
|---|---|---|
| empty set → `0.0` | red (`an_empty_assertion_map_scores_one_not_zero`, and `assertions::tests::the_score_is_the_passing_fraction`) | red (`test_an_empty_..._not_zero`) |
| fraction → always `1.0` | red (3 tests) | red (2 tests) |
| `assertion_map` keeps the **first** value | red (`a_repeated_name_...`) | red (same test) |
| list path stops delegating (empty list → `0.0`) | red (`the_list_and_the_map_score_the_same_assertions_alike`) | red (same test, `case=[]`) |

The list/map agreement test is the one I would have expected to be decorative and it is
not: it is the thing that catches the two shapes drifting apart, and it caught it.

**CI covers the empty case cross-language already — better than the PR body says.** The PR
declines to add a conformance corpus and justifies it. Checking the justification: the
committed corpus has six `score/*` cases including **`score/no-assertions`**, and after this
change both probes route those through the shared rule — the Python probe via
`score_from_assertions` (`conformance/probe_assertions.py:194`) and Rust via
`assertions::score` (`differential_assertions.rs:353`). So the empty-set answer is now
asserted byte-identical across languages by the existing differential harness, which was not
true of `result::score_from` as a separate copy. The PR undersells this.

**Behaviour really is unchanged by the unification.** The refactor introduces a
`passed == total` short-circuit into Rust's `assertions::score`, which previously always
divided. `n/n` is exactly `1.0` in IEEE-754, so this is a no-op — and the 47k sweep plus
the differential harness reproducing the committed corpus confirm it empirically rather
than by argument.

**The `ruff format` disclosure checks out.** `ruff format --check` fails on `models.py` on
the default branch too; `ruff check` passes on the PR. Pre-existing, correctly not touched.

## Outside this PR

*Neither of these is this PR's job and neither should affect the verdict.*

- **The edit to a sibling change's changelog entry.** `[Unreleased] → Rust — Changed` opened
  with a bullet reading "Nothing.", which this PR made false and rewrote to "On SVG
  geometry, nothing." The author flagged it. It is the right fix and the smallest one, but
  it is a structural symptom: a "Nothing." bullet in a shared `[Unreleased]` section is a
  claim about the whole section that any concurrent change invalidates. Worth a convention
  ("Nothing." only survives to release time, never sits in `[Unreleased]`), not a change
  here.
- **Neither `score_from` nor `assertion_map` is exported from the Python package root.** The
  PR is right that this matches `score_from_assertions`, which is not either, and right to
  leave it. If the maintainer wants them at the root it should be one change touching both
  implementations, as the PR suggests.

PR #201, opened 2026-08-20
recipe: RecipeMeta, the descriptive half of a recipe without the declarative runner

### LINKED ISSUE #199: Recipe metadata is only reachable through the declarative recipe model
Same shape as the `Selectable`/`Comparable` gap closed in 0.4.0: a useful data
structure that only callers of the declarative runner can reach.

`recipe::Recipe` declares `name`, `description`, `intent`, `priority`,
`execution`, `determinism`, `ci_paths`. A consumer whose recipes are imperative
— they branch on what the screen shows, so they cannot be expressed as a
declarative `Recipe`, which the docs correctly call out as out of scope — has to
re-declare all seven in its own metadata struct. Measured: ~90 lines, and the
field names match one for one.

That consumer's other nine fields are genuinely product-specific
(`enable_tool_calls`, `yolo`, `timeout`, `permissions`, `renderers`, ...) — and
`Recipe` already carries `extension: HashMap<String, serde_json::Value>`, which
is the sanctioned home for exactly that. So the entire struct could be
upstream's, with the product half in `extension`.

**Suggestion.** Split the descriptive fields into a `RecipeMeta` (or similar)
that `Recipe` embeds and an imperative consumer can construct directly. This
also makes `Selectable` almost free for such consumers, since `name` and
`ci_paths` would then come from a type the crate owns.

To be clear about scope: **not** asking for the imperative recipe model to be
supported. Driving the session directly is the consumer's job and the
declarative/imperative split is the right one. This is only about the metadata
that describes a recipe, which is the same either way.

md-mt:
Closes #199.

`recipe::Recipe` declared `name`, `description`, `intent`, `priority`, `execution`, `determinism` and `ci_paths` itself. A consumer whose recipes are imperative cannot construct a `Recipe` — it needs a `command` such a suite has no single honest value for — so it re-declared all seven, field for field. Those seven describe a scenario the same way whichever kind it is, so they are now `recipe::RecipeMeta`, which `Recipe` holds and flattens.

No imperative runner was added, and no product-specific field. `Recipe::extension` remains the sanctioned home for consumer knobs.

## `Recipe`'s serialised form and schema are unchanged — demonstrated

I dumped the serialised output and the generated schema from the **pre-split tree**, made the change, and diffed:

| Artifact | Result |
|---|---|
| A fully-populated recipe (every field non-default, extension fields at recipe and command level) | byte-identical |
| A recipe that takes every default | byte-identical |
| `schema::generate_recipe_schema()` | byte-identical |

Those pre-split strings are now checked in as assertions in `rust/crates/termproof/tests/recipe_meta.rs`, so the pin is against the old shape rather than a fresh recording of the new one. The schema is covered by the existing `schema_snapshot.rs`, which needed no re-blessing — and, mutated, still reports drift *through* the flattened type (`/properties/priority/description`), so it is guarding rather than merely passing.

Key order matters and is preserved: `meta` sits where `name` used to, so the flattened keys emit in exactly their old positions. Two flattened fields on one struct is the arrangement where a known key could fall through to the catch-all, so there is an explicit test that `extension` still holds only `my_extra`.

## `Recipe`'s Rust source API does change — and `cargo semver-checks` is right

```
Checked 196 checks: 194 pass, 2 fail
  constructible_struct_adds_field: field Recipe.meta
  struct_pub_field_missing:        name, description, intent, priority, execution, determinism, ci_paths
Summary semver requires new major version
```

**This is a genuine public-API break, not a false positive.** `recipe.name` becomes `recipe.meta.name`, and a `Recipe { .. }` literal has to supply `meta`. There is no way in Rust to have both a flat named field and a nested one; embedding is what the issue asks for and this is its price. Under the project's pre-1.0 rule that is the minor digit — **I have not bumped anything**; the version consequence is yours.

I considered `impl Deref<Target = RecipeMeta> for Recipe`, which would keep field *reads and writes* compiling. Declined: it saves neither the struct literals nor `semver-checks`, so it buys partial compatibility at the price of a smart-pointer conversion on a plain data type, and it would hide the break rather than state it. Say the word if you'd rather have it.

Blast radius inside this repository was **three test assertions and one trait impl** — `recipe::Recipe` has far fewer callers than a grep for `.name` suggests, because `runner`, `execution`, `agent` and the CLI all use the *other* `Recipe` (see "Noticed, not fixed").

## `Selectable` does get easier — it falls out

`Selectable` is now implemented on `RecipeMeta`, and `Recipe`'s impl delegates to it. An imperative consumer that holds a `RecipeMeta` passes it straight to `select`/`select_names` with no trait impl of its own — the issue's "almost free" is free. There is a doctest on `RecipeMeta` showing exactly that, and a unit test that builds metadata with no `command` in sight.

## Python: mirrored in substance, not in mechanism

`models.RecipeMeta` is the same seven fields with the same defaults, and `Recipe.meta` hands over a recipe's descriptive half (copying `ci_paths` rather than aliasing it). Two deliberate differences, both stated in the code:

**`Recipe` does not inherit from `RecipeMeta`.** That would be the closer mirror, but dataclass inheritance puts base fields first, which moves `command` behind six defaulted fields and forces it keyword-only — changing `Recipe`'s positional signature. This package states it avoids exactly that break, in `StepResult`'s docstring ("positional or keyword — is unaffected"), so I kept `Recipe` flat and untouched and made `meta` a property. The cost is that the seven names and defaults are declared twice with nothing in the language holding them together, so `tests/test_recipe_meta.py` holds them instead, field by field. `Recipe`'s field list, order, positional signature and the absence of a `meta` *field* are all pinned there too — those pins are the reason the inheritance option can be rejected on evidence rather than taste.

**Selection needed nothing.** `selection.select_names` has always taken `(name, ci_paths)` pairs, so the gap `Selectable` closed on the Rust side in 0.4.0 never existed in Python. `RecipeMeta` satisfies `select_recipes` structurally too. I did not widen its annotation to a Protocol — that is a Python-only API addition the issue does not ask for.

## Conformance harnesses: not extended, and why

The three differential harnesses compare *runtime behaviour* over a checked-in corpus (steps, assertions, evidence manifest). A metadata default table is a type surface, not a behaviour, and the canonical schema declares no `default` for any of the seven, so there is nothing in the shared corpus for it to hang off. Instead the default table is asserted verbatim on both sides under the same test name — `recipe_meta.rs::test_defaults_are_the_shared_table` and `test_recipe_meta.py::test_defaults_are_the_shared_table` — each pointing at the other. Happy to build a real harness for cross-implementation *type* surfaces if you want one; that felt like a separate change.

I also paired `termproof.models` with the new test in `run_stdlib_tests.py`'s `STDLIB_ONLY` map (it had `None`), so the new surface is held to the stdlib-only claim too. That job passes.

## Tests: every new test was broken on purpose

Each mutation was applied, the suite run, then reverted; the tree is green at HEAD.

| Mutation | Went red |
|---|---|
| Drop `#[serde(flatten)]` on `Recipe.meta` | 6 tests |
| Move `meta` to the end of `Recipe` (key order) | both serialised-text tests |
| `RecipeMeta::default().priority` → `"P9"` | defaults table + parsed-equals-constructed |
| Drop `#[serde(default)]` on `determinism` | 3 tests |
| `RecipeMeta::new` stops using the serde defaults | parsed-equals-constructed |
| `Selectable for Recipe` returns a constant | delegation test |
| `Selectable for RecipeMeta` returns `&[]` | 2 tests + the `RecipeMeta` doctest |
| A doc comment on a `RecipeMeta` field | `schema_snapshot.rs` |
| PY: `RecipeMeta.priority` default drifts from `Recipe`'s | defaults table + field-agreement |
| PY: `meta` aliases `ci_paths` instead of copying | copy test |
| PY: `meta` declared as a field | 3 tests |
| PY: `command` made keyword-only | positional-signature test |
| PY: the property mis-maps `intent` | hand-over test |

**Reported honestly:** two tests are not covered by a subtle mutation. `test_metadata_is_constructible_without_a_command` and `test_metadata_selects_the_same_way_a_recipe_does` fail only on a coarse break (the type stops existing, or gains a required field — which fails at import). They document the seam rather than guard a subtle regression, and I have said so rather than inventing a mutation for them.

**One flaky full-suite run.** Of six `unittest discover` runs over `python/tests`, five were `OK (834 tests)` and one reported 1 failure + 2 errors. I did not capture the names before it went green again and could not reproduce it in four subsequent runs. My change adds a frozen dataclass and a property; it cannot produce intermittent failures. Flagging it rather than omitting it.

## Validation

- `cargo test --workspace --all-features` — green
- `cargo clippy --workspace --all-features --all-targets` — clean
- `cargo fmt --all --check` — clean
- `python -m unittest discover -s tests` — 834 green (see flake note)
- `python scripts/run_stdlib_tests.py` — 154 green
- `ruff check` — clean; `mypy` on the two touched/added Python files — clean

`ruff format --check` reports `python/termproof/models.py` as unformatted. **Pre-existing** — I verified it fails identically on `origin/main`, at two spots neither of which I touched. Left alone.

## Noticed, not fixed

- **Rust has two `Recipe` types.** `recipe::Recipe` (typed, schema-generating, extension-preserving) is what `schema`, `selection` and `validation` mean; `models::Recipe` (also exported as `ModelRecipe`) is a narrower struct with no `determinism`, `ci_paths`, `renderers` or extension map, and it is what `runner`, `execution`, `agent` and the CLI actually run. Python has one. This issue is about `recipe::Recipe` only, so `models::Recipe` is untouched — but it is why the internal blast radius was three assertions, and it is a real trap for anyone changing "the recipe model". I added a short note to `AGENTS.md` rather than acting on it.
- **Python `Recipe` has no `extension` map**, so the issue's "put the product half in `extension`" advice is Rust-only. Pre-existing and already documented in `recipe.rs`'s module docs.


md-mt:
Rebased onto `5d34f1d` (#200) and addressed the review. Force-pushed as `39eb301`.

## Rebase

Resolved as the reviewer described: the only conflict was textual in `CHANGELOG.md`, where #200 added the scoring bullet and renamed `- Nothing. \`SvgMetrics\` …` to `- On SVG geometry, nothing. …`. Kept every side — #200's scoring bullet, #200's renamed SVG bullet verbatim, and this PR's bullet between them. No code conflict. Re-validated on the rebased tree, not on the pre-rebase one.

## The semver-checks break is waived, deliberately, and narrowly

Per the captain's decision: **termproof stays on 0.4.x and this break ships waived, not fixed.** No version number moved.

The waiver is **by finding, not by lint, and the job still runs.** `cargo-semver-checks` 0.50.0 has no per-item waiver — no `--exclude-lint`, no JSON output to post-process, and I confirmed empirically that `[package.metadata.semver-checks.lints]` is *not* read at this version (the findings still fail with it set). The mechanisms it does offer are "silence a lint crate-wide" and "drop the job", and both would waive breaks nobody has looked at.

So `.github/scripts/rust/semver-checks.py` wraps the tool. Every lint still runs against the same published baseline and the tool's own output is echoed verbatim; the only addition is holding the findings to the eight named entries in `rust/semver-waivers.toml` (which sits beside `deny.toml`, the same kind of config in the same place):

```toml
waived = [
  "constructible_struct_adds_field: field Recipe.meta",
  "struct_pub_field_missing: field name of struct Recipe",
  ... the six others ...
]
```

`field ci_paths of struct Recipe` is waived. `field checks of struct Recipe` is not — a future change removing it fails this job with the lint still saying so.

**It fails in both directions, and the second one is the point.** An unwaived finding fails, so a different break cannot ride along on the waiver written for this one. A waiver that matches *nothing* also fails — so when you bump the version and publish, the baseline moves past these breaks, the entries stop matching, and CI asks for the waiver to be deleted. A waiver that cannot expire is a disabled check with extra steps.

Verified end to end against real `cargo semver-checks` output, not just unit tests:

| Scenario | Result |
|---|---|
| As committed | exit 0, `8 finding(s), all waived`, reason echoed |
| One waiver entry removed | exit 1, names `field ci_paths of struct Recipe` as unwaived |
| A waiver entry that matches nothing | exit 1, names the stale entry and says to delete it |
| Build broken so the tool reports nothing | exit 1 — not silently green |

The waiver file also requires a `reason`, and `test_semver_waivers.py` asserts the `[Unreleased]` changelog actually describes the break, so a waiver cannot be added without a consumer-facing note. The CHANGELOG's `**What breaks:**` paragraph is unchanged and now carries a `**The version did not move.**` paragraph saying the break is waived and where.

New tests: `python/tests/test_semver_waivers.py` (9), following the existing `test_version_bump.py` pattern of testing a `.github/scripts/rust/*.py` from the Python suite.

## Review findings

**#1 — crate-root export: taken.** `pub use recipe::{… Recipe, RecipeMeta, Step …}`. Your reasoning is the part I'd missed: the population this break lands on is exactly the population editing an import line already, and asking them to also discover `termproof::recipe::` is a second papercut. It also left the two front doors asymmetric, since the Python side does export from the package root. Now pinned on both sides — `recipe_meta.rs::the_type_is_reachable_from_the_crate_root` and `test_recipe_meta.py::test_the_type_is_reachable_from_the_package_root`, each naming the other. Both go red when the export is removed (Rust: `E0433`; Python: one mutation for `__all__`, one for the import).

That Python test needed a guard I'd have missed: `test_recipe_meta.py` is wired into `run_stdlib_tests.py`, where `termproof` is a namespace-only stand-in with no `__all__`. It `skipTest`s there with the reason spelled out, per the convention in `CLAUDE.md`. Confirmed skipping under the gate (155 tests, 7 skipped) and running in the full suite.

**#2 — the aliasing asymmetry: taken, docs only.** You're right that only one side was documented and that the changelog described the copy as if it were neutral. Rust's `meta` is an owned field, so `recipe.meta.ci_paths.push(x)` mutates the recipe; Python's is a property returning a copy, so the same expression is a no-op. Both doc comments now say so and name the other implementation. No behaviour changed.

**#3 — "free" was overstated: agreed, correcting it here.** Your distinction is the right one. A consumer holding a bare `RecipeMeta` gets selection with zero lines. A consumer holding `struct ImperativeRecipe { meta: RecipeMeta, yolo: bool, … }` — which is the issue's actual worked example — still writes a four-line `impl Selectable` delegating to `self.meta`, because `select` is generic over `R: Selectable` and hands back `Vec<&R>`. So: four delegating lines instead of seven re-declared fields, which is a real improvement and is what the issue's "almost free" meant. I've left the code alone; the claim was the problem.

**Two corrections to my own earlier report,** both from your run rather than mine:

- I said `test_metadata_is_constructible_without_a_command` only fails on a coarse break. You found drifting `RecipeMeta.determinism`'s default trips it. It guards more than I claimed.
- Two mutations I hadn't thought of — reordering `intent`/`priority` inside `RecipeMeta`, and adding an eighth field — both go red. Good to know the net is tighter than I'd characterised it.

On the Python flake: still unreproduced, now across your two full runs and my six. Neither of us has cleared it and I'm not treating clean runs as a verdict.

## Validation on the rebased tree

- `cargo test --workspace --all-features` — green
- `cargo clippy --workspace --all-features --all-targets` — clean; `cargo fmt --all --check` — clean
- `python -m unittest discover -s tests` — 849 green
- `python scripts/run_stdlib_tests.py` — 155 green, 7 skipped
- `ruff check` clean; `mypy` clean on the three touched/added Python files
- `.github/scripts/rust/semver-checks.py` — exit 0, all 8 findings waived, plus the four scenarios in the table above
- Workflow YAML parses; the step resolves correctly under the job's `working-directory: rust`, and `rust/semver-waivers.toml` and `.github/scripts/rust/**` are both already inside this workflow's path filters

Your "Outside this PR" items I've left alone: the two `Recipe` types (noted in `AGENTS.md`, worth its own issue), the unpinned Python export list (this PR pins one name, not the mechanism), and `name`'s `minLength: 1` laxity, which predates this change.

PR #202, opened 2026-08-20
cli: the clap layer that turns a command line into a RunConfig, behind an opt-in feature

### LINKED ISSUE #197: run_config models the whole CLI surface, but nothing populates it from a command line
`run_config` is a complete model of how a run is configured — `RunConfig`,
`Discovery`, `Selection`, `Execution`, `BinarySource`, `Output`, `Publisher`,
`Requirements`, plus `parse`, `from_path`, and the `pick(flag, configured,
builtin)` precedence helper. Everything a caller needs to *describe* a run.

Then every Rust consumer hand-writes the argument parsing that fills it in.

Measured on one consumer (a TUI validator, 15 imperative recipes): `cli.rs` is
1,763 lines with **23 clap argument builders and 37 functions**. These exist
only to get from flags to a `RunConfig` the crate already defines:
`parse_args`, `args_from_matches`, `explicit`, `explicit_flag`,
`apply_configured_selection`, `apply_configured_binary`, `selection_config`,
`selection_config_for`, `setting`, `check_publishers`. Roughly 200 lines, and
the `pick()` precedence rules get reimplemented per flag by hand.

The Python implementation ships `termproof.cli` (326 lines). Rust has no `cli`
module, so this looks like an omission rather than a deliberate asymmetry.

**Suggestion.** A `clap`-shaped builder that emits the standard flags for
`RunConfig` and parses them back into one — something like
`run_config::clap_command()` / `run_config::from_matches(&ArgMatches)` — behind
a feature flag so the dependency is opt-in. A consumer with extra flags of its
own can `.arg()` onto the returned command; a consumer with none gets the whole
outer loop free.

Not asking for a binary. Asking for the layer that produces the config type the
crate already owns.

md-mt:
Closes #197.

`run_config` has modelled a whole run since it landed — `RunConfig`, `Discovery`,
`Selection`, `Execution`, `BinarySource`, `Output`, `Publisher`, `Requirements`,
and `pick(flag, configured, builtin)` for how the three sources of an answer
rank. Nothing populated it from argv, so each Rust consumer wrote that layer
again. This adds it once, behind an opt-in feature.

## The shape

`run_config::cli`, gated on the new default-off `clap` feature, re-exported at
the `run_config` root:

| Function | What it is |
|---|---|
| `clap_command() -> Command` | the standard flags on a bare `termproof` command |
| `augment_args(Command) -> Command` | the same flags added to a command the consumer owns |
| `from_matches(&ArgMatches)` | exactly what the flags said, everything else unset |
| `configured(&ArgMatches)` | the file `--run-config` names, or an empty config |
| `merge(flags, configured, builtin)` | `pick()` over every field, in one place |
| `resolve(&ArgMatches, builtin)` | all of the above in one call |

The issue suggested `clap_command()` / `from_matches(&ArgMatches)`; both exist
with those names. `resolve` is the addition, and it is the one the issue's
complaint actually asks for: `from_matches` alone gives you the flag column, and
a caller left to combine that with a config and a built-in is a caller
re-deriving `pick()` per flag, which is the 200 lines. `augment_args` is the
composability seam — `clap_command()` is defined as `augment_args` over a bare
command, so the two cannot drift.

## Two rules that are asserted, not described

**No flag carries a `default_value`.** `run_config`'s own precedence section
already warned about this: a flag that merely *has* a default is
indistinguishable from one the caller *passed*, and if the two look alike a
config file can never override it, which would make most of the schema dead. So
the built-in arrives as the `builtin: &RunConfig` argument, and
`no_flag_carries_a_default_value` walks `clap_command().get_arguments()` and
fails on any default — structural, so a later flag cannot quietly reintroduce
the bug.

**A flag passed empty is a flag passed.** `--renderer ''` beats a configured
renderer and resolves to `Some("")`. Reading it as unset would add a fourth
state on top of `pick()`'s three and put every consumer back to remembering
which flags treat empty specially. Covered directly by
`a_flag_passed_empty_is_still_a_flag_passed`.

**And one field does not obey that precedence.** `Requirements.uploaded_media`
is a `bool`, so it has no unset state for `pick()` to rank and `merge` ORs the
three layers. It can be raised but never lowered: `require: {uploaded_media:
false}` in a config file is inert against a built-in `true`, and there is no
`--no-require-uploaded-media`. For a requirement that is the safe direction to
fail in, but it is an exception rather than an instance of the rule, so it is
now stated in the module doc and the changelog and pinned by
`a_required_upload_can_be_raised_but_never_lowered`.

For the fields that are a `Vec` or a map rather than an `Option`, empty *is*
unset and the first non-empty of flag / config / built-in wins as a whole. Two
consequences are documented rather than left to be discovered: there is no way
to say "explicitly no excludes" on the command line, which matches the config
file where `exclude: []` is also indistinguishable from an absent key; and a
passed `--env` replaces a configured `env` wholesale rather than merging key by
key, because per-key merging is a fourth precedence rule and this layer exists
to stop those being invented one consumer at a time.

## Reconciled against `termproof.cli`

Where the two name the same thing they use the same flag: `--priority`,
`--recipe-name` (repeatable in both), `--renderer`. Every divergence:

| This layer | `termproof.cli` | Why |
|---|---|---|
| `--run-config PATH` | `--config PATH` is the *plugin registry* (`VerifierConfig`) | Two different config files. Reusing `--config` is the one name collision worth avoiding, and a consumer with both still needs to name each. |
| `--artifact-dir DIR` | `--out DIR` | Not the same thing. `--out` is one directory the report path is also derived from; `RunConfig` separates `artifact_dir`, `report_path` and `result_json_path`, so no flag here carries `--out`'s meaning and taking the name would be exactly the collision above. |
| `--priority` **with** `--recipe-name` is a usage error | the two compose | The one divergence that is not about a name. `termproof.registry.select_recipes` applies priority then names as successive filters, so `--priority P0 --recipe-name smoke` is a legal narrowing there — "smoke, if it is P0". `Selection` holds one arm, so the same pair is an `ArgumentConflict` here. Pinned by `priority_and_recipe_name_do_not_compose_here_though_they_do_in_python`. |
| `--all`, `--changed-files` | — | `Selection` has four arms and Python's CLI exposes two. All four get a flag, in one `ArgGroup`, so "all *and* a priority" is a usage error rather than something this layer has to rank. |
| `--root`, `--repo-marker`, `--exclude`, `--transport`, `--model`, `--effort`, `--binary-installed`, `--binary-build`, `--env`, `--timeout-scale`, `--publisher`, `--publisher-setting`, `--require-uploaded-media`, `--require-media-publisher` | — | Fields `RunConfig` has and Python has no counterpart for. |

`--binary-installed` / `--binary-build CHANGE` in one `ArgGroup` mirror
`BinarySource`'s two arms the way the selection group mirrors `Selection`'s
four: mutually exclusive by construction on both sides, so neither the enum nor
the command line can hold two and there is no validation pass to forget.

## Why Python is unchanged

`run_config` is Rust-only — there is no `termproof.run_config`, and
`conformance/README.md`'s three harnesses are steps, assertions and the evidence
manifest. There is no cross-implementation surface here to extend. This closes
an asymmetry (Rust had no `cli` module at all) rather than opening one, and
`termproof.cli` is an `argparse` *program* where this is a library layer. No
binary is added.

## Feature gating and CI

`clap = ["dep:clap"]`, off by default — the only feature that is. The reasoning
is in `Cargo.toml` and the README: the other four *subtract* from a shape that
was already published, so default-on is what keeps turning one off from being a
break; this one *adds* a shape that never was, so default-off is what keeps it
from being a change.

CI's feature powerset is enumerated from the `FEATURES` array, so adding `clap`
grew it from sixteen combinations to thirty-two by construction. Both halves are
covered: sixteen combinations without `clap`, sixteen with. I ran all 32 locally
(clippy `-D warnings` + `cargo test`) and every one passes.

Both halves are covered: sixteen combinations without `clap`, sixteen with. I
ran all 32 locally (clippy `-D warnings` + `cargo test`) and every one passes.

**On the cost of that doubling.** Measured on run 32429616067, the last `main`
run before this branch: the powerset step is 7.8 min on ubuntu and 9.3 on macos
for sixteen combinations — ~30-35s each — in a job totalling 10.1 and 12.1.
Thirty-two puts the step at ~16-19 min and the job at ~18-21, inside the
existing 45-minute bound, so `timeout-minutes` is unchanged. The workflow
comment now carries those numbers and notes that a *sixth* feature would be
~37 min of powerset in a ~40 min job — past that headroom and near the
60-minute per-job cap `python/tests/test_ci_timeouts.py` enforces — which is
where sharding across the job matrix becomes necessary.

## Verification

- Full Python suite: `uv run python -m unittest discover -s tests` — 830 tests, all pass.
- `cargo fmt --check --all`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace` — all clean.
- All 32 feature combinations: clippy + test, all pass.
- `cargo deny check` — advisories, bans, licenses, sources all ok. The lockfile
  gains one line (`clap` under the `termproof` entry); `termproof-cli` already
  depended on it, so no new version enters the graph.
- `cargo semver-checks check-release -p termproof` — 196 checks, 196 pass, "no
  semver update required". Not a public-API break: the new items are behind a
  feature that is off by default, so the compared surface is unchanged. No
  version number is touched; the CHANGELOG entry is under `[Unreleased]`.
- Every one of the 23 new tests was broken deliberately and confirmed to go red.
  Reported below, because one of them did not the first time.

## Two corrections to this PR as first pushed

The first push raised the `rust` job's `timeout-minutes` from 45 to 75 to absorb
the doubled powerset, and that turned all three Python unit-test jobs and
`Build, verify TUI evidence and publish it` red — one cause, four jobs.
`python/tests/test_ci_timeouts.py` caps every job in every workflow at 60
minutes, deliberately: a timeout there is a backstop, not a budget, and the
point is to fail in minutes rather than shave an hour off GitHub's six-hour
ceiling (#183). The cap was right and the raise was wrong — I extrapolated from
a warm-cache laptop figure of ~50s per combination instead of reading the CI
history recorded in the docstring of the very test I was breaking. Reverted to
45 in `dfc5e3e`, with the measured numbers written into the comment.

## Review findings addressed

All six minors from the review, plus the nit. Findings 4 and 2 were flagged as
the issue's whole point; both are now disclosed *and* pinned by a test, since
the failure mode for a disclosure is that it silently stops being true.

| # | Finding | What changed |
|---|---|---|
| 4 | Boolean requirements escape `pick()` and it was undisclosed | Stated in the module doc and the changelog entry: `uploaded_media` is a `bool`, so it is ORed and can be raised but never lowered. Behaviour left alone, as the reviewer suggested — a config file quietly switching off an evidence gate is the outcome worth preventing. New test `a_required_upload_can_be_raised_but_never_lowered` pins both halves, including that no lowering flag exists. |
| 2 | `--priority` + `--recipe-name` compose in Python, conflict here, undisclosed | Verified against `registry.py:78-89` — they are successive filters there. Row added to the divergence table in the changelog and in this body; a paragraph in the module doc explains it is the one divergence not visible in a flag *name*. New test `priority_and_recipe_name_do_not_compose_here_though_they_do_in_python`. |
| 5 | Three doc spots now false | Fixed all three: `lib.rs`'s feature section ("first four", plus a `clap` bullet), `engineering-baseline.md` (sixteen → thirty-two, plus where the doubling ceiling is), `CONTRIBUTING.md` (count and list, plus a note that `cargo test --workspace` does not compile this module). |
| 6 | Three new rustdoc warnings in the default build | Fixed by dropping the link syntax, per the reviewer's second option. `cargo doc` is back to 6 warnings on default features and 6 on `--all-features`, matching `main`. |
| 1 | `clap` drags in `derive` | Taken, with a different remedy — see below. |
| 3 | Reserved long names undocumented, collision panics | Taken. The `augment_args` doc now lists all 23 long names in order, the argument ids, and both `ArgGroup` names, and says the collision is a debug-only panic. |
| 7 | `from_matches` panics on foreign matches (nit) | Taken, same doc block: the matches must come from a command this module augmented, and the subcommand path is named as the realistic way in. |

### Finding 1: took the finding, changed the remedy

The reviewer is right that `derive` is dead weight — nothing in the workspace has
a `#[derive(Parser)]`, including `termproof-cli`, which says so in its own module
doc. Dropping it from the workspace pin removes `clap_derive` and `heck` from the
lockfile entirely, and `termproof-cli` stops paying for it too.

Two corrections to the finding's arithmetic and its suggested fix:

- It removes **two** crates, not four. `syn`, `quote` and `proc-macro2` stay
  regardless, because `serde` and `thiserror` need them — measured with
  `cargo tree -p termproof --no-default-features --features clap`: 80 crates
  before, 78 after. One of the two is a proc macro, which is the part worth
  removing.
- I did **not** take `default-features = false, features = ["std"]`. That removes
  nine crates rather than two, and I checked what the other seven buy before
  giving them up. Under `std`-only, `render_help()` returns the empty string —
  every `.help()` string in this module is silently discarded — and errors
  degrade from `the argument '--all' cannot be used with '--priority <PRIORITY>'`
  to `an argument cannot be used with one or more of the other specified
  arguments`. This layer's entire product is a `Command` a consumer calls
  `get_matches()` on; shipping one with a blank `--help` to save seven crates is
  the wrong trade, and a consumer cannot re-enable it from their side. The
  reasoning and both measurements are in the `Cargo.toml` comment.

The README's cost column now reads "`clap`, builder-only (no `clap_derive`)".

## The test that stayed green

`a_consumer_can_add_its_own_flags` survived a mutation that made `augment_args`
throw away the `Command` it was handed and build a fresh one — the test asked
only that the consumer's flag and ours both parse, which they still did, since
`clap` largely ignores `argv[0]`. That is a real hole in the composability
claim: a consumer's name, about, version and subcommands would have been
silently discarded. The test now asserts `command.get_name() == "validator"`,
and goes red under both that mutation and the more obvious "augment_args adds
nothing". Every other test went red on the first attempt.

## Noted, not fixed

`merge` and its `pick_opt` / `pick_list` / `pick_map` helpers are pure and have
no `clap` in them, so they could live in `run_config` proper and serve a
consumer with a hand-rolled parser. They are inside the feature gate here to
keep the default build's public surface untouched; lifting them out is additive
and belongs in its own change.

Six pre-existing `rustdoc` warnings in the crate (`last_error`, `cell_width`,
two on `BUILTIN_TYPES`, two redundant explicit links) are unrelated to this diff
and left alone.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -33,7 +33,12 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 45
+    # 75, not 45, because the feature powerset doubled with `clap` (#197) and
+    # that step is most of the job. Measured locally on a warm cache at roughly
+    # 50 seconds per combination, so thirty-two is ~28 minutes before the other
+    # steps and before a cold CI build. The headroom is for the doubling, not a
+    # sign the job got slower per unit of work.
+    timeout-minutes: 75
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: dtolnay/rust-toolchain@6c977a6ca4077a0ceb28ffbe03f59d46e9ac8772 # master (2026-08-05)
@@ -49,21 +54,25 @@ jobs:
       # `--all-features` cannot catch a feature that only builds alongside
       # another, which is the failure mode an undeclared `dep:` produces. So
       # check the whole powerset rather than a sample (#27, extended for
-      # `schema` in #28 and for `junit` in #34). `json-schema` implies `schema`,
-      # so some combinations resolve to the same set as others — they are run
-      # anyway, because what is being checked is that every combination a
-      # consumer can *write* compiles, not every set cargo can resolve.
-      # `default` and `--all-features` are the same set today and the steps
-      # above already cover them.
+      # `schema` in #28, `junit` in #34 and `clap` in #197). `json-schema`
+      # implies `schema`, so some combinations resolve to the same set as
+      # others — they are run anyway, because what is being checked is that
+      # every combination a consumer can *write* compiles, not every set cargo
+      # can resolve. `clap` is the first feature that is not in `default`, so
+      # `default` and `--all-features` are no longer the same set: the steps
+      # above cover `--all-features`, and the combination naming the other four
+      # below is `default`.
       #
       # The list is enumerated from FEATURES rather than written out, so adding
       # a feature grows the matrix by construction and cannot be half-extended:
-      # four features is sixteen combinations, and the count is printed so a
-      # silently shrinking matrix is visible in the log.
+      # five features is thirty-two combinations, and the count is printed so a
+      # silently shrinking matrix is visible in the log. Growth is doubling, so
+      # a sixth feature is the point at which this wants splitting across the
+      # job matrix rather than another entry in the array.
       - name: clippy and test each feature combination
         run: |
           set -eux
-          FEATURES=(evidence junit json-schema schema)
+          FEATURES=(evidence junit json-schema schema clap)
           total=$((1 << ${#FEATURES[@]}))
           echo "feature powerset: $total combinations"
           for mask in $(seq 0 $((total - 1))); do
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -33,12 +33,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
-    # 75, not 45, because the feature powerset doubled with `clap` (#197) and
-    # that step is most of the job. Measured locally on a warm cache at roughly
-    # 50 seconds per combination, so thirty-two is ~28 minutes before the other
-    # steps and before a cold CI build. The headroom is for the doubling, not a
-    # sign the job got slower per unit of work.
-    timeout-minutes: 75
+    timeout-minutes: 45
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: dtolnay/rust-toolchain@6c977a6ca4077a0ceb28ffbe03f59d46e9ac8772 # master (2026-08-05)
@@ -66,9 +61,18 @@ jobs:
       # The list is enumerated from FEATURES rather than written out, so adding
       # a feature grows the matrix by construction and cannot be half-extended:
       # five features is thirty-two combinations, and the count is printed so a
-      # silently shrinking matrix is visible in the log. Growth is doubling, so
-      # a sixth feature is the point at which this wants splitting across the
-      # job matrix rather than another entry in the array.
+      # silently shrinking matrix is visible in the log.
+      #
+      # Growth is doubling, so it is worth writing down where the ceiling is.
+      # Measured on run 32429616067, the last `main` run before `clap`: this
+      # step was 7.8 min on ubuntu and 9.3 on macos for sixteen combinations —
+      # ~30-35s each — in a job totalling 10.1 and 12.1. Thirty-two puts the
+      # step at ~16-19 min and the job at ~18-21, which is why the job's
+      # `timeout-minutes` is unchanged at 45. A sixth feature would be ~37 min
+      # of powerset in a ~40 min job, past that headroom and near the 60-minute
+      # cap `python/tests/test_ci_timeouts.py` enforces on every job. That is
+      # the point at which this wants sharding across the job matrix rather
+      # than another entry in the array.
       - name: clippy and test each feature combination
         run: |
           set -eux
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -328,6 +328,55 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   sub-module path would be a second papercut on top of the first.
   ([#199](https://github.com/md-mt/termproof/issues/199))
 
+- **`run_config::cli`, behind the new opt-in `clap` feature:** the command line
+  that fills in a `RunConfig`. `run_config` has modelled a whole run since it
+  landed — `RunConfig`, `Discovery`, `Selection`, `Execution`, `BinarySource`,
+  `Output`, `Publisher`, `Requirements`, and `pick(flag, configured, builtin)`
+  for how the three sources of an answer rank — but nothing populated it from
+  argv, so each consumer wrote that layer again. Measured on one of them, a TUI
+  validator: 23 `clap` argument builders and about 200 lines whose only job was
+  to get from flags to this crate's own type, with `pick()` re-derived by hand
+  once per flag (#197).
+
+  Six functions, re-exported at the `run_config` root:
+  `clap_command()` for the standard flags on a bare command;
+  `augment_args(Command)` for adding them to a consumer's own, so extra `.arg()`
+  calls compose and the consumer keeps its name, version and subcommands;
+  `from_matches(&ArgMatches)` for exactly what the flags said;
+  `configured(&ArgMatches)` for the file `--run-config` names;
+  `merge(flags, configured, builtin)` for `pick()` applied to every field at
+  once; and `resolve(&ArgMatches, builtin)` for all of it in one call.
+
+  Two rules are load-bearing and are asserted rather than described. No flag
+  carries a `default_value` — a defaulted flag is indistinguishable from a
+  passed one, and if the two look alike a config file can never override one,
+  which is what `run_config`'s precedence section warned about; the built-in
+  therefore arrives as the `builtin: &RunConfig` argument. And a flag passed
+  empty is a flag passed: `--renderer ''` beats a configured renderer, because
+  reading it as unset would add a fourth state to `pick()`'s three and put every
+  consumer back to remembering which flags have it.
+
+  **Reconciled against `termproof.cli`.** Where the two name the same thing they
+  use the same flag: `--priority`, `--recipe-name` (repeatable in both) and
+  `--renderer`. The rest is new, and no name is reused for a second meaning:
+
+  | This layer | Python's `termproof.cli` | Why |
+  |---|---|---|
+  | `--run-config PATH` | `--config PATH` is the *plugin registry* (`VerifierConfig`) | Two different config files. Reusing `--config` would be the one collision worth avoiding, and a consumer that has both still needs to name each. |
+  | `--artifact-dir DIR` | `--out DIR` | Not the same thing. `--out` is one directory that the report path is also derived from; `RunConfig` separates `artifact_dir`, `report_path` and `result_json_path`, so no flag here has `--out`'s meaning. |
+  | `--all`, `--changed-files` | — | `Selection` has four arms; Python's CLI exposes two of them. All four get a flag, in one `ArgGroup`, so "all *and* a priority" is a usage error rather than something this layer has to rank. |
+  | `--root`, `--repo-marker`, `--exclude`, `--transport`, `--model`, `--effort`, `--binary-installed`, `--binary-build`, `--env`, `--timeout-scale`, `--publisher`, `--publisher-setting`, `--require-uploaded-media`, `--require-media-publisher` | — | Fields `RunConfig` has and Python has no counterpart for. |
+
+  The feature is off by default, so a consumer that does not name it does not
+  compile `clap`. CI's feature powerset is enumerated from the feature list, so
+  it grew from sixteen combinations to thirty-two by construction, and both the
+  default build without `clap` and every build that names it are exercised.
+
+  Python is deliberately unchanged. `run_config` is Rust-only — there is no
+  `termproof.run_config`, and the conformance pair has nothing to compare —
+  so this closes an asymmetry rather than opening one. `termproof.cli` is an
+  `argparse` *program*; this is a library layer, and no binary is added.
+
 ### Rust — Changed
 
 - **scoring:** `assertions::score` and `RunResult::score_from_assertions` were
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -356,6 +356,20 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   reading it as unset would add a fourth state to `pick()`'s three and put every
   consumer back to remembering which flags have it.
 
+  **One field does not obey that precedence, and it is deliberate.**
+  `Requirements.uploaded_media` is a `bool`, so it has no unset state for
+  `pick()` to rank and the three layers are ORed instead: any of them asking for
+  it is asking for it. It can therefore be raised but never lowered —
+  `require: {uploaded_media: false}` in a config file is inert against a
+  built-in `true`, and there is no `--no-require-uploaded-media`. For a
+  requirement that is the safe direction to fail in, since a config file quietly
+  switching off a "the evidence must be shareable" gate is the outcome worth
+  preventing, but it is an exception to the rule rather than an instance of it:
+  a caller that needs the gate off has to leave it off in the built-in. The
+  repeatable flags behave differently again — for those, empty is unset, and the
+  first non-empty of flag / config / built-in wins as a whole rather than
+  merging, so a passed `--env` replaces a configured `env` outright.
+
   **Reconciled against `termproof.cli`.** Where the two name the same thing they
   use the same flag: `--priority`, `--recipe-name` (repeatable in both) and
   `--renderer`. The rest is new, and no name is reused for a second meaning:
@@ -365,6 +379,7 @@ with the pre-1.0 rule that under `0.x` a breaking change bumps the minor digit.
   | `--run-config PATH` | `--config PATH` is the *plugin registry* (`VerifierConfig`) | Two different config files. Reusing `--config` would be the one collision worth avoiding, and a consumer that has both still needs to name each. |
   | `--artifact-dir DIR` | `--out DIR` | Not the same thing. `--out` is one directory that the report path is also derived from; `RunConfig` separates `artifact_dir`, `report_path` and `result_json_path`, so no flag here has `--out`'s meaning. |
   | `--all`, `--changed-files` | — | `Selection` has four arms; Python's CLI exposes two of them. All four get a flag, in one `ArgGroup`, so "all *and* a priority" is a usage error rather than something this layer has to rank. |
+  | `--priority` **with** `--recipe-name` is a usage error | the two compose | The one divergence that is not about a name. `termproof.registry.select_recipes` applies priority and then names as successive filters, so `--priority P0 --recipe-name smoke` is a legal narrowing there — "smoke, if it is P0". `Selection` holds one arm, so the same pair is an `ArgumentConflict` here. Same names, same individual meanings, different composition, which is exactly the case a ported command line trips over; a caller wanting the intersection filters a `Selection::Priority` run itself. |
   | `--root`, `--repo-marker`, `--exclude`, `--transport`, `--model`, `--effort`, `--binary-installed`, `--binary-build`, `--env`, `--timeout-scale`, `--publisher`, `--publisher-setting`, `--require-uploaded-media`, `--require-media-publisher` | — | Fields `RunConfig` has and Python has no counterpart for. |
 
   The feature is off by default, so a consumer that does not name it does not
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -197,10 +197,12 @@ cargo test  --manifest-path rust/Cargo.toml --workspace
 ```
 
 Run all three before pushing. CI also runs the suite across the whole feature
-powerset (four features — `evidence`, `junit`, `json-schema`, `schema`, so
-sixteen combinations) and at the declared dependency floors, so a change that
-only breaks a non-default combination is caught there even if your local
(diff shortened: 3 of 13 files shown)


md-mt:
# PR 202 — cli: the clap layer that turns a command line into a RunConfig, behind an opt-in feature

**VERDICT:** MERGE WITH NITS
**MERGE RISK:** low — nothing here changes behaviour for an existing consumer; the residue is one dependency that is heavier than it needs to be and four documentation spots that now say something false.
**CI:** 18 passed, 0 failed, 1 skipped of 19. Meaningful coverage for this change: the feature powerset runs 16 combinations without `clap` and 16 with, on two operating systems, `--all-features` clippy and test compile the new module, and the Python job runs the workflow-contract lint that the first push tripped.
**STALENESS:** current. `pr-202` has current `main` as an ancestor and `git merge-tree` reports zero conflicts. I re-fetched the head after finishing (`dfc5e3e`, unchanged), so nothing below was judged against a stale tree.

## Summary

This does what the issue asked and does it carefully: a library-only `clap` layer, opt-in, that emits the standard `RunConfig` flags and parses them back with `pick()` applied once. I mutation-tested the two load-bearing claims rather than reading them, and both hold. The measured CI numbers in the workflow comment are exact to the second. The findings below are all small; the largest is that the opt-in dependency pulls a proc-macro chain the module never uses, which quietly contradicts the reason the feature is opt-in.

## Findings

### 1. The `clap` dependency drags in `derive`, which this module never uses — minor
**Location:** `rust/crates/termproof/Cargo.toml:119` (`clap = { workspace = true, optional = true }`), inheriting `rust/Cargo.toml:236` (`clap = { version = "4.5", features = ["derive"] }`)

**Verified.** `cli.rs` is entirely builder API — `Arg::new`, `ArgGroup`, `value_parser!`. There is no `#[derive(Parser)]` anywhere in it. But the optional dependency inherits the workspace's `features = ["derive"]`, so turning the feature on compiles a proc-macro toolchain for nothing:

```
$ cargo tree -p termproof --features clap -e normal | grep -E 'clap|syn |quote|proc-macro2'
├── clap v4.6.5
│   ├── clap_builder v4.6.5
│   │   ├── clap_lex v1.1.0
│   └── clap_derive v4.6.4 (proc-macro)
│       ├── proc-macro2 v1.0.107
│       ├── quote v1.0.47
│       └── syn v3.0.3
```

Failure case: a consumer follows the README, adds `features = ["clap"]` to get one `Command` builder, and pays for `clap_derive`, `syn`, `quote` and `proc-macro2` — three of which are compile-time-only crates that cannot be shared with a build that has no other proc-macro user. The whole justification for this feature being off by default, stated in `Cargo.toml` and the README, is that a consumer should not compile a layer it will not call. The dependency as declared undercuts that. The README's cost column also understates it: it says the feature costs "`clap`".

**What I would do:** `clap = { workspace = true, optional = true, default-features = false, features = ["std"] }`. One line. Worth doing in this PR, since the feature's whole argument is about what a consumer pays for.

### 2. `--priority` and `--recipe-name` compose in the Python CLI and conflict here, and the PR's divergence table does not say so — minor
**Location:** `rust/crates/termproof/src/run_config/cli.rs:151-156` (the `select` `ArgGroup`), against `python/termproof/registry.py:78-89`

**Verified.** In `termproof.cli`, `select_recipes` applies `priority` and then `names` as successive filters, so `--priority P0 --recipe-name smoke` is a legal narrowing — "the recipe named smoke, if it is P0". Here the two are in one `ArgGroup(multiple = false)`:

```
$ # probe against the built crate
3 priority+recipe-name -> Some(ArgumentConflict)
```

The PR body's reconciliation table lists `--priority` and `--recipe-name` in the "where the two name the same thing they use the same flag" row, and separately explains that all four `Selection` arms live in one `ArgGroup`. It never joins those two statements up. The result is that two flags the table presents as *matching* behave differently in composition: legal in one implementation, a usage error in the other. That is the "same name, different meaning" case the table exists to rule out, and it is the divergence most likely to bite someone porting a command line across.

This follows from `Selection` being an enum, which predates the PR — so it is not a defect in the code. It is a gap in a disclosure the PR explicitly claims is complete ("Every divergence").

**What I would do:** add the row. One line in the table saying the two are mutually exclusive here and composable there.

### 3. The module does not say which long option names it claims, and a collision panics — minor
**Location:** `rust/crates/termproof/src/run_config/cli.rs:110-270`

**Verified.** `augment_args` adds 22 long options, several of them generic enough that a consumer plausibly already owns one: `--root`, `--model`, `--exclude`, `--env`, `--publisher`, `--transport`, `--effort`. A collision is not an error, it is a panic:

```
thread 'main' panicked at clap_builder/src/builder/debug_asserts.rs:110:17:
Command x: Long option names must be unique for each argument, but '--root'
is in use by both 'root' and 'myroot'
```

Two things make this worth a doc line rather than nothing. First, `clap`'s `debug_assert` machinery only runs in debug builds, so a consumer that only ever builds release gets duplicate long names resolved by `clap`'s internal ordering rather than a diagnostic. Second, the composability seam is the headline feature of this layer, and the doc comment for `augment_args` — "The consumer keeps its own `Command` … and any flags of its own" — reads as if there were nothing to check.

**What I would do:** list the reserved long names (or at least the argument ids) in the `augment_args` doc comment, so a consumer can check before it panics.

### 4. Boolean requirements escape `pick()` entirely, and the PR's rules section does not mention it — minor
**Location:** `rust/crates/termproof/src/run_config/cli.rs:397-402`

**Verified.** `Requirements.uploaded_media` is a `bool`, so `merge` ORs the three layers instead of ranking them. A built-in that sets it cannot be lowered by anything:

```
7 builtin-on, no flag -> true
```

That includes the config file: `require: { uploaded_media: false }` in a `--run-config` file is inert against a built-in `true`. There is no `--no-require-uploaded-media`. For a *requirement* this is a defensible fail-safe, and the code comment at that line says exactly why. The gap is disclosure: the PR body's "two rules that are asserted, not described" section presents flag-beats-config-beats-builtin as the whole story, and neither the README nor the changelog entry mentions that one field does not obey it. A consumer that reads the changelog and writes `uploaded_media: false` expecting it to take effect will get silence.

**What I would do:** one sentence in the changelog entry and in the module doc, alongside the existing paragraph about lists and maps. The behaviour itself I would leave.

### 5. Three documentation spots now say something false — minor
**Verified** by grep and read; all three were correct before this PR and are wrong after it.

- `rust/crates/termproof/src/lib.rs:39` — the crate-root rustdoc still reads "All four are on by default, so a consumer that does not name features gets the whole crate", and its bullet list of features omits `clap`. There are now five features, `--all-features` is no longer the same set as `default`, and the crate root is the first place a consumer looks. The PR updated the README, the architecture doc and the status-and-parity doc, which carry the same list, but not this one.
- `rust/docs/engineering-baseline.md:189` — "With four features that is the full powerset of sixteen". This is the file that `Cargo.toml`'s own feature comment cites as the policy authority for feature design, so it is the exact document the next person adding a feature will read.
- `CONTRIBUTING.md:200-201` — "CI also runs the suite across the whole feature powerset (four features — `evidence`, `junit`, `json-schema`, `schema`, so sixteen combinations)". Stale count and stale list; a contributor reading this will not learn that `clap` exists.

**What I would do:** update all three. About six lines total, and the PR is otherwise meticulous about keeping these in step.

### 6. Three new rustdoc warnings in the default build — minor
**Location:** `rust/crates/termproof/src/run_config.rs:20-23`

**Verified.** The new paragraph adds intra-doc links to ``[`cli`]``, ``[`clap_command`]`` and ``[`resolve`]``, but the paragraph itself is not feature-gated while the items are:

```
$ cargo doc -p termproof --no-deps            # default features
warning: unresolved link to `cli`
warning: unresolved link to `clap_command`
warning: unresolved link to `resolve`
warning: `termproof` (lib doc) generated 9 warnings

$ cargo doc -p termproof --no-deps --all-features
... 6 warnings
```

Six on `main`, nine here. The PR body accounts for "six pre-existing rustdoc warnings … unrelated to this diff and left alone" while adding three more of exactly that kind, unmentioned. Published docs are unaffected, because the crate sets `all-features = true` for the doc build; this only shows up for a consumer running `cargo doc` against a default build.

**What I would do:** wrap the paragraph in `#[cfg_attr(feature = "clap", doc = "...")]`, or drop the link syntax and name the items in plain text.

### 7. `from_matches` returns a `Result` but panics on foreign matches — nit
**Location:** `rust/crates/termproof/src/run_config/cli.rs:466`

**Verified.** Handed an `ArgMatches` from a command that was never augmented, it panics rather than returning `Err`:

```
thread 'main' panicked at run_config/cli.rs:466:10:
Mismatch between definition and access of `root`. Unknown argument or group id.
```

This is standard `clap` behaviour and arguably plain misuse, but there is one realistic path into it: a consumer that calls `augment_args` on a *subcommand* and then passes the top-level `ArgMatches`. Given that the signature advertises a recoverable error, one doc line saying the matches must come from a command this module augmented would earn its keep.

## What is good

**The `pick()` precedence is genuinely enforced, not merely asserted.** I mutation-tested it rather than reading the tests:

| Mutation | Tests that went red |
|---|---|
| `.default_value("default")` added to `--renderer` | 4, including the structural `no_flag_carries_a_default_value` |
| `pick_opt` reordered so configured beats flag | 3, including `a_flag_passed_empty_is_still_a_flag_passed` |
| `pick_list` / `pick_map` reordered the same way | 2 |

The structural test does catch the bug it claims to catch. The empty-flag case is real and correctly ranked. And the built-in-only field (`transport` in `a_flag_beats_the_config_which_beats_the_builtin`) is the half a hand-rolled implementation usually drops, and it is asserted.

**Composability works, including the part the author found broken and fixed.** I wrote the consumer the issue describes and ran it:

```
1 watch=true sel=Some(Priority("P0"))
1 subcommand survives: ["doctor"]
2 renderer=Some("alt") transport=Some("pty")
```

A consumer's name, version, about, subcommands and own flags all survive `augment_args`, and `resolve` then applies precedence against the consumer's built-in correctly. The self-reported near-miss — a test that stayed green under a mutation that discarded the consumer's `Command` — is honestly reported and genuinely closed by the `get_name()` assertion.

**The feature gate is real, not nominal.** `cargo tree -p termproof -e normal` returns zero `clap` hits with default features and zero with `--no-default-features`; four only with `--features clap`. Both halves of the gate are exercised in CI by construction, since the powerset is enumerated from the `FEATURES` array. I spot-checked `""`, `clap`, `clap,junit` and `clap,json-schema,schema` locally with clippy `-D warnings` and test — all clean — and `cargo fmt --check` and `clippy --all-features -D warnings` pass on the branch.

**The CI timing claim is exact.** This is the claim I most expected to be a rounded guess, and it is not. Against the job API for the cited run:

| | powerset step | claimed | job total | claimed |
|---|---|---|---|---|
| ubuntu | 7 m 48 s | 7.8 min | 10 m 07 s | 10.1 min |
| macos | 9 m 17 s | 9.3 min | 12 m 09 s | 12.1 min |

And this PR's own run validates the extrapolation rather than leaving it as arithmetic: the powerset step came in at 15.2 min on ubuntu and 16.9 on macos, with job totals of 17.5 and 19.6 — at or just under the predicted 16-19 and 18-21, against an unchanged 45-minute bound. Keeping the timeout at 45 is right, with roughly 2.3x headroom. `python/tests/test_ci_timeouts.py` passes on the branch.

**It stayed a library layer.** No `[[bin]]`, no `src/bin/`, no new binary target — only `pub mod cli` behind the gate. The issue's "not asking for a binary" is respected.

**The declared dependency floor is honest.** The floors CI step runs default features, so it never compiles this module, and I expected the `clap = "4.5"` floor to be aspirational. It is not: pinning the workspace to `=4.5.0` and checking `-p termproof --no-default-features --features clap` compiles clean.

## On why a "Rust-only" change broke the Python suite

Worth stating plainly, because the framing invites a wrong conclusion. **The change was not Rust-only.** The first push also edited `.github/workflows/rust-ci.yml`, raising the `rust` job's `timeout-minutes` from 45 to 75.

The coupling is deliberate, documented and working as designed:

- `python/tests/test_ci_timeouts.py` is a repository-wide workflow-contract lint that happens to live in the Python suite. It parses every `.github/workflows/*.yml` and caps every job at 60 minutes, with a docstring explaining why the cap is a backstop rather than a budget.
- `.github/workflows/python-ci.yml` lists `.github/workflows/**` in its trigger paths, with a 20-line comment explaining that this suite is where the workflow contracts are asserted and that eight of nine such tests used to sit outside the filter.
- Four jobs went red from one cause because the TUI evidence job also runs the unit tests (`python-ci.yml:271`), which the PR body states correctly.

So this is not a hidden coupling and not a finding against the design — the guard fired exactly where it was built to fire, on the first push, before a human looked. What it is, is a process point: the author skipped the local Python suite on a change that edited a workflow file, and the repository's own `CONTRIBUTING.md` gates section does not say that editing `.github/workflows/**` requires running the Python suite. That is the one durable gap, and it is small.

The PR body's account of the failure matches what I found in the fix commit and the test. Nothing was papered over.

## Outside this PR

- The six pre-existing rustdoc warnings the PR body names (`last_error`, `cell_width`, two on `BUILTIN_TYPES`, two redundant explicit links) are real and confirmed on `main`. Leaving them alone is the right call for this diff; they want their own change.
- `CONTRIBUTING.md`'s gates sections list the commands to run per language but never say that a change to `.github/workflows/**` is asserted by the Python suite. Adding that one line would have prevented this PR's red push, and will prevent the next one. Not this PR's job, but cheap.
- The author's own "noted, not fixed" — that `merge`, `pick_opt`, `pick_list` and `pick_map` are pure and have no `clap` in them, so they could serve a consumer with a hand-rolled parser from `run_config` proper — is a good observation and correctly deferred. Lifting them out is additive and belongs in its own change.

## What I could not verify

Nothing material. I could not read the full CI log for the reference run (the log endpoint returns only the tail), so the per-step timings above come from the job API's `started_at`/`completed_at` rather than from log timestamps; the two would not disagree.

PR #203, opened 2026-08-21
release: a hand-pushed rs-v tag publishes its own crates

### LINKED ISSUE #191: Rust release workflow cannot create its own GitHub release, so crates.io never publishes
The Rust release workflow **attaches** assets to a GitHub Release it expects to
already exist. Nothing creates that release. The Python workflow creates its own.

Observed on 0.4.0: pushing `rs-v0.4.0` built all three target binaries
successfully, then failed at `Attach to release → Wait for the release to exist`.

**It cascades.** `rust-publish-crates.yml` triggers on `release: published`. No
release means no crates.io publish, so the Rust half of the release was tagged
and built but entirely unpublished until a release was created by hand.

For 0.4.0 the recovery was: create the release manually, re-run the failed job,
and the crates publish then fired on its own. That worked, but it is a manual
step nobody is told about, and the failure mode is quiet — the binaries build
green and the tag exists, so the release looks like it happened.

## The fix

Make `rust-release.yml` create the release if it does not exist, the way
`python-release.yml` does. The two workflows should be symmetric; there is no
reason for one to require a prerequisite the other creates for itself.

## Worth checking while there

- **Does the wait step have a purpose?** If it was written for a flow where
  something else created the release first, that flow no longer exists.
- **Should a failed attach fail the release loudly?** It did fail here, which is
  good — but the *consequence* (no crates.io publish) was silent, because that
  is a separate workflow that simply never triggered.
- **Is there a check that the tag, the release, and both registries agree after
  a cut?** Every part of 0.4.0 succeeded individually while the release as a
  whole was half-published. A post-release verification step would have caught
  it in seconds.

That last point is the more valuable one. This class of bug — every component
green, the composition broken — is the same shape as several defects found in
this repository recently.

md-mt:
Closes #191.

## The gap

`rs-v0.4.1` was pushed by hand. `rust-release.yml` built the three binaries and then waited for a GitHub release that nothing on the tag-push path was going to create, waited 150 seconds, and failed with *"no GitHub release exists for tag 'rs-v0.4.1' … Create the release, then re-run this job."*

That reads as a problem attaching binaries. It was not one. `rust-publish-crates.yml` triggers on a **published release** and nothing else, so crates.io received nothing at all — and one red job among several green ones was the entire signal that a release had silently not happened. It was recovered by hand: creating the release fired the crates publish and let the attach job succeed on re-run.

Python never had this problem, though not by sharing this shape: `python-release.yml` creates its own release from the tag *and* uploads to PyPI as a later step of the same job, gated on the tag ref. Its tag path is self-sufficient because everything happens in one job. Rust's indirection through a release event is deliberate and is what makes the `crates-io` environment gate possible — so the fix is to make the tag path produce that release, not to collapse the two designs.

## What changed

**`rust-release.yml` — the attach job creates the release when the wait finds none.**
The existing wait loop is what makes this safe rather than something added beside it: it now records `exists=true|false` instead of ending in `exit 1`, and the create step is gated on that output. On the weekly path the release is created seconds after the tag push and this job is minutes behind, so the first check finds it and nothing is created — still exactly one release. By the time the loop has failed ten times over 150s, nothing else is going to create it.

The step must use `RELEASE_TOKEN`, and **its absence fails the job rather than falling back**. This is the crux: a release created with the default `GITHUB_TOKEN` does not start `rust-publish-crates.yml`, and because a `workflow_dispatch` run there is a dry run by design, no re-run could ever publish it. A tag with no release is recoverable; a release that can never publish is not. It also re-checks the tag against `publish-plan.py`'s version *before* creating anything, so it cannot mint a release the publish workflow would refuse.

**Notes come from the same renderer the weekly path uses.** `release-decide.py` grows one flag, `--at VERSION`: describe the range instead of deciding on it. Without it the script derives the *next* version from the commits, so notes for a tag that already exists would be headed one release too far ahead with a compare link to a tag nobody cut — locally, `--from rs-v0.4.0 --to rs-v0.4.1` decides `0.5.0`. Nothing else about the deciding mode moves.

**The path ends by asking crates.io.** A new `verify-crates` job runs `.github/scripts/rust/verify-published.sh`, which polls the sparse index for every crate `publish-plan.py` names at this version and fails naming the ones that are not there. It runs on both release paths. The failure mode worth designing against is not a red job — it is a release page that looks finished beside a registry that received nothing, and no amount of reading the run list makes that obvious. A follow-up step reports what the `Publish crates (Rust)` runs did, because "never ran" and "ran and failed" look identical from the registry and only the first is invisible.

## Constraints

- **No double release.** Creation is gated on the wait's output, and the wait outlasts the automation's own create by two orders of magnitude. Asserted in `test_the_release_is_created_only_when_the_wait_found_none`, including step ordering.
- **The publish trigger is unchanged.** `release: published`, no `push:`, same `if:` on the publish job, same `crates-io` environment. Asserted in `CratesPublishStaysTheOneGateTest`.
- **Nothing else uploads to a registry.** Asserted: `rust-release.yml` contains no `cargo publish` and no `CARGO_REGISTRY_TOKEN`.
- **No crate is named.** The verifier is asserted against the actual crate directory names; the set is derived from `publish-plan.py`, like everywhere else.
- **Notes are not a regression.** Rendered by `release-decide.py`, asserted non-empty and sectioned.

## How I verified the tag path without cutting a release

Four layers, all offline or read-only.

**1. Ran the create step's actual shell, with `gh` stubbed.** `CreateStepShellTest` lifts the step's `run:` block out of the YAML and executes it under `bash -e` — what GitHub uses — against a fake `gh` that records its arguments. Six branches:

| scenario | result |
|---|---|
| happy path | `gh release create rs-v0.4.1 --title rs-v0.4.1 --notes-file release-notes.md --verify-tag`, exit 0 |
| `RELEASE_TOKEN` empty | exit 1, the error, **zero `gh` calls** |
| tag disagrees with manifest | exit 1, **zero `gh` calls** |
| create loses a race, release exists | exit 0, `::notice::`, no failure |
| create genuinely fails | exit 1 |
| notes on disk | start `## termproof 0.4.1` |

The tag is derived from the manifest rather than written down, so it tests the guard and not a version.

**2. Rendered the notes 0.4.1 should have had.** Running the workflow's own command against the real `rs-v0.4.0..rs-v0.4.1` range produces the correct heading, the conventional-commit sections, and `compare/rs-v0.4.0...rs-v0.4.1`:

```
## termproof 0.4.1

### Breaking changes

- **svg:** one source of truth for the SVG geometry defaults, on a Linux-safe font stack (#195) (`e4e3de2f`)
...
**Full changelog:** https://github.com/md-mt/termproof/compare/rs-v0.4.0...rs-v0.4.1
```

**3. Ran the verifier against a local `file://` sparse index.** `curl` reads `file://` like `https://`, so `VerifyPublishedScriptTest` builds the crates.io shard layout in a tmpdir and exercises sharding, version matching, polling, and the empty-plan case with no network and no dependency on what is published. Separately, against real crates.io: the live plan set at `0.4.1` passes; `99.99.99` and a held-back crate both fail naming exactly what is missing.

**4. Mutation-tested the workflow assertions.** Dropping the wait guard, swapping `secrets.RELEASE_TOKEN` for `github.token`, and deleting the `verify-crates` job each turn the suite red (1 error, 1 failure, 4 errors). The tests are not vacuous.

Also: full Python suite green (885 tests), `check_version.py`, `check_schema_copies.py`, `run_stdlib_tests.py` (159) all pass, `shellcheck` clean on the new script.

## Not done here

`publish.sh` and `verify-published.sh` each carry their own eight-line sparse-index URL helper. Factoring it into a shared file would be a structural change to the publish path in a behavioural diff; worth doing, separately.

## Follow-up in this PR

CI caught a real bug in the first commit, which is the payoff for testing the step's shell rather than only reading the YAML. The step resolved revisions through `$TAG^{commit}`. That is the same commit HEAD already is — the tag push started the run — but it is a name that has to have been fetched, so in a clone without tags `rev-parse` failed outright instead of falling through to the whole-history branch. `5bba742` moves both revisions to `HEAD` / `HEAD^`, verified against a `git clone --no-tags --depth 1`, where the fallback now renders notes and the release is created.

CI on the final commit: 19 passed, 1 skipped — the skip being `Publish`, which is exactly the property that the publish job still runs only on a published release.

## Review round (`0fa3154`)

**Finding 1 — the registry check was skipped exactly when it mattered.** Correct and the sharpest finding here. `needs: attach` carries an implicit `success()`, so a release created followed by a failed upload — the same shape as the incident — left a red *Attach to release* beside a registry nobody looked at. Now `if: ${{ !cancelled() && startsWith(github.ref, 'refs/tags/rs-v') }}`.

That makes the job reachable after a failed build, where 40 minutes of polling buys nothing, so it asks the cheap question first: a new step checks the release exists and fails in seconds if not — naming the consequence, *a tag went out and shipped nothing*, rather than waiting out a publish that was never triggered. Both mutations (reverting the guard, deleting the fast-fail step) now turn the suite red; neither did before.

**Finding 2 — dispatch is no longer always a dry run at the system level.** Also correct. Said plainly in `rust/docs/publishing.md` and in `rust-release.yml`'s header: dispatching *Release (Rust)* against a tag is the one hand-trigger that can lead to a publish, deliberately, as the recovery action for a tag whose release is missing — bounded by the tag ref, the manifest match, the absence of a release, and the token. Fixing this also moved the dry-run paragraph back under the heading it describes; an earlier insertion in this branch had orphaned it under "Confirming the crates actually shipped".

**Finding 3 — taken.** Two assertions pinning `rust-release.yml`'s triggers and the tag-ref guards on `attach` and `verify-crates`. Both of your mutations were silent before and fail now. Your reading is right that the version guard inside the create step is what carries the safety; this makes the outer layer deliberate rather than incidental.

**Finding 4 — declined, with a reason.** A hard `args.at != current` exit would break the flag's stated purpose. `--at` describes a release that already exists, and the useful hand invocation is regenerating notes for an *older* tag from a current checkout — where `cargo metadata` reads the working tree's manifest and the two versions legitimately differ. There is also no surprise to guard against: the caller passed the version explicitly and gets notes headed at it. The one caller that must not diverge already checks, before `gh` is invoked.

**Finding 5 — declined, with a reason.** Agreed on the smell, but every fix I costed is worse than the problem. A `git worktree` writes into the shared git common directory and leaves a stale registration if a test dies; a clone per test is slow and loses the history the notes range needs; running with `cwd=tmp` breaks the relative script and manifest paths the step legitimately uses. Changing the workflow to write elsewhere would be changing production for a test. The `setUp` skip plus `addCleanup` covers the realistic case, and as you say the current form is honest about running the real thing.

**Finding 6 — fixed, in the PR body above.** You are right and the correction matters in the direction you name: PyPI is uploaded by a step in the same job, gated on the tag ref, not by a release event. Rust's indirection through `release: published` is the thing that makes the environment gate possible, so the two paths are self-sufficient for different reasons. The CHANGELOG entry makes no claim about Python, so nothing there needed changing.

**Outside this PR, noted:** `"Nothing else in this repository uploads to a registry"` is loose as written — `python-release.yml` uploads to PyPI. Pre-existing, plainly meant as "to crates.io", and left alone.

### DIFF
--- a/.github/scripts/rust/release-decide.py
+++ b/.github/scripts/rust/release-decide.py
@@ -9,6 +9,10 @@ scheduled workflow runs can be run by hand against any range:
     .github/scripts/rust/release-decide.py --from A --to B # any range
     .github/scripts/rust/release-decide.py --to B          # first-release path
 
+`--at` turns the same command into a description of a release that already
+exists rather than a decision about one — see the flag's help. Everything below
+is about the deciding mode.
+
 The two questions it answers are independent, and both must come back yes.
 
 1. Is there a change worth releasing?
@@ -285,6 +289,14 @@ def main():
         action="store_true",
         help="release even when nothing releasable changed; the version rule still applies",
     )
+    parser.add_argument(
+        "--at",
+        metavar="VERSION",
+        help="the version being released is exactly this — describe the range "
+        "instead of deciding on it. For a tag that already exists, where the "
+        "version is a fact rather than something a bump derives; the notes and "
+        "the reported tag name that version.",
+    )
     args = parser.parse_args()
 
     head = git("rev-parse", args.head).strip()
@@ -309,7 +321,18 @@ def main():
 
     first_release = base is None
 
-    if not commits and not args.force:
+    if args.at:
+        # Describing, not deciding. The caller already has a tag, so asking
+        # whether to release is moot and deriving a bump would name the version
+        # *after* the one that went out. A quiet range is fine here — a release
+        # cut by hand is allowed to be a release nothing would have proposed.
+        release, reason, level, nxt = (
+            True,
+            f"describing the release already tagged at {args.at}; no bump was derived",
+            "none",
+            args.at,
+        )
+    elif not commits and not args.force:
         release, reason, level, nxt = (
             False,
             "no commits since the last release other than the automation's own version bump",
--- a/.github/scripts/rust/verify-published.sh
+++ b/.github/scripts/rust/verify-published.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+#
+# Confirm that every crate the publish plan names is on crates.io at VERSION.
+#
+# Reads ORDER (space-separated crate names) and VERSION from the environment,
+# the same contract publish.sh has, so the set being checked is the set that
+# was meant to be uploaded. Nothing here names a crate; the caller derives both
+# values from .github/scripts/rust/publish-plan.py.
+#
+# ## Why this exists as its own thing
+#
+# The publish workflow already waits for each crate it uploads to appear on the
+# index, so a *publish that ran* reports its own failures. The gap this closes
+# is the publish that never ran at all: it triggers on a published GitHub
+# release, so a tag that produced no release produces no upload and no red job
+# — only the absence of one, which reads as nothing having gone wrong.
+#
+# Absence is not something the publish workflow can report on. This runs on the
+# tag, at the end of the release path, and asks the registry the only question
+# that matters: is the thing the plan said would ship actually there.
+#
+# ## Polling, and what a timeout means
+#
+# The crates upload is downstream of the release this path publishes, and it
+# runs its own gate first, so it finishes minutes after the binaries do. Hence
+# a poll rather than a single look. A timeout is reported as a failure and
+# names the crates that are missing: at that point either the publish never
+# fired, it failed, or it is still waiting — and all three are states a release
+# should not be left in silently. If the `crates-io` environment ever grows a
+# required reviewer, an approval slower than this window will land here; raise
+# ATTEMPTS rather than deleting the check.
+#
+# ATTEMPTS and INTERVAL are overridable so the same script can be exercised
+# without waiting out the release-sized window.
+set -euo pipefail
+
+: "${ORDER?ORDER must be set to the space-separated publish order (may be empty)}"
+: "${VERSION:?VERSION must be set to the workspace version}"
+
+ATTEMPTS="${ATTEMPTS:-80}"
+INTERVAL="${INTERVAL:-30}"
+INDEX="${INDEX:-https://index.crates.io}"
+
+# Sparse-index layout: crates.io shards by name length, then by prefix.
+index_url() {
+    local name=$1
+    case ${#name} in
+        1) printf '%s/1/%s\n' "$INDEX" "$name" ;;
+        2) printf '%s/2/%s\n' "$INDEX" "$name" ;;
+        3) printf '%s/3/%s/%s\n' "$INDEX" "${name:0:1}" "$name" ;;
+        *) printf '%s/%s/%s/%s\n' "$INDEX" "${name:0:2}" "${name:2:2}" "$name" ;;
+    esac
+}
+
+# The index is what a consumer's `cargo build` resolves against, so it — not
+# the web API — is the thing worth asking. A 404 means the crate has never been
+# published at any version.
+on_index() {
+    local name=$1 body
+    body=$(curl -sSf --max-time 30 "$(index_url "$name")" 2>/dev/null) || return 1
+    jq -e -s --arg v "$VERSION" 'map(.vers) | index($v) != null' >/dev/null <<<"$body"
+}
+
+if [ -z "${ORDER// /}" ]; then
+    # Every member carries `publish = false`. That is a policy recorded in the
+    # manifests, not a fault, and the publish workflow would upload nothing
+    # either — so there is nothing to be missing.
+    echo "::notice::the publish plan names no crates, so there is nothing to verify on crates.io"
+    exit 0
+fi
+
+missing=""
+for ((attempt = 1; attempt <= ATTEMPTS; attempt++)); do
+    missing=""
+    for crate in $ORDER; do
+        on_index "$crate" || missing="$missing $crate"
+    done
+    if [ -z "${missing// /}" ]; then
+        echo "every planned crate is on crates.io at $VERSION: $ORDER"
+        exit 0
+    fi
+    echo "waiting for crates.io ($attempt of $ATTEMPTS) — still missing at $VERSION:$missing"
+    if [ "$attempt" -lt "$ATTEMPTS" ]; then
+        sleep "$INTERVAL"
+    fi
+done
+
+echo "::error::the release is out but these crates are not on crates.io at $VERSION:$missing — the crates publish did not happen. It triggers on a published GitHub release and nothing else, so check that the release exists and was published by an identity whose events start workflows, then look at the Publish crates (Rust) runs."
+exit 1
--- a/.github/workflows/rust-publish-crates.yml
+++ b/.github/workflows/rust-publish-crates.yml
@@ -13,6 +13,14 @@
 # failed partway, re-run this workflow from the release; the publish step is
 # idempotent and will skip what already landed.
 #
+# That single gate is deliberate and is what makes the `crates-io` environment
+# approval hook below possible, so nothing here reaches for the tag instead.
+# The cost is that a tag with no release publishes nothing, silently. Both
+# paths that produce a tag therefore also produce a release —
+# rust-auto-release.yml creates both, and rust-release.yml creates the release
+# when a hand-pushed tag arrives without one — and rust-release.yml ends by
+# checking crates.io for the crates this workflow was supposed to upload.
+#
 # ## Why the upload is its own job
 #
 # The upload runs in the `crates-io` GitHub environment. Entering an
--- a/.github/workflows/rust-release.yml
+++ b/.github/workflows/rust-release.yml
@@ -61,6 +61,15 @@
 # the fact. A missing token therefore fails this job rather than creating a
 # release that can never publish.
 #
+# One consequence is worth stating plainly, because publish-crates.yml's header
+# says that workflow cannot be hand-triggered into publishing and a reader will
+# generalise it. Dispatching *this* workflow against an `rs-v*` tag takes the
+# same path a tag push does, so it can create a release, and a release is what
+# publishes. That is the intended way to recover a tag whose release is
+# missing, and it is bounded by the same guards: a tag ref, a manifest at
+# exactly that version, no release already there, and the token. A dispatch on
+# a branch reaches none of it — the attach job is gated on the ref being a tag.
+#
 # ## The check at the end
 #
 # Everything above still only proves that the release *path* ran. The failure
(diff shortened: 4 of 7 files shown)


md-mt:
# PR 203 — release: a hand-pushed `rs-v` tag publishes its own crates

**VERDICT:** MERGE WITH NITS
**MERGE RISK:** low — the change is confined to the tag-push path of one workflow, is gated by the existing wait, and cannot widen what publishes; the residual risk is that the new registry check does not run when an upstream job in the same workflow fails.
**CI:** 19 passed, 1 skipped, 0 failed. Meaningful for this change: the full Python suite (3 interpreter versions) executes the new 27-test module, and the one skip is `Publish` — which is itself the property under test, i.e. a pull request still cannot reach the upload job.
**STALENESS:** Not stale. `origin/main` (`49dbaf6`) is an ancestor of the PR head (`5bba742`); verified with `git merge-base --is-ancestor`. I re-fetched `refs/pull/203/head` at the end of the review and it is still `5bba742`.

## Summary

The fix is correct and the trace closes. A hand-pushed `rs-v*` tag now completes the whole path: tag push → build → wait finds no release → create it with a token whose events start workflows → `release: published` → the crates publish → a new job asks the registry whether the crates are actually there. I traced every link, ran the new create step's real shell in a tagless shallow clone, ran the new verifier against both a synthetic index and the live registry, and independently mutation-tested the assertions. The publish trigger is genuinely untouched — the only change to the publish workflow is eight lines of comment.

The findings below are all minor. None of them is a reason to hold the change.

## Findings

### 1. minor — the registry check does not run when anything upstream fails
`.github/workflows/rust-release.yml:344-347`

`verify-crates` has `needs: attach` and `if: startsWith(github.ref, 'refs/tags/rs-v')`. That `if:` contains no status function, so the default `success()` on `needs` still applies and the job is **skipped** whenever `attach` fails or is skipped. Verified by reading; the semantics are GitHub's documented `needs` behaviour, which I could not execute here.

Concrete failure case: the release is created (the new step succeeds), then `Attach the archives and checksums` fails — a transient upload error is enough. The release is published, the crates publish has already been triggered and may or may not have landed, and the run shows a red job named *Attach to release*. `verify-crates` never runs. That is the same shape as the incident this PR exists for: a red job that reads as an attachment problem while the registry state is unknown. The difference — and it is a real one — is that re-running the job now produces a release and a publish, whereas before it produced nothing, so the natural response works. That is why this is minor rather than major.

The same applies when a build target fails: `attach` is skipped, so nothing says the tag shipped nothing. That case is a deliberate design choice documented at `rust-release.yml:191-195`, and is out of scope here.

What I would do: `if: !cancelled() && startsWith(github.ref, 'refs/tags/rs-v')`. `always()` is already the house pattern in this repository (`python-release.yml:122,140`, `python-ci.yml:427`). The cost is that a failed build would add ~40 minutes of polling before a second red job; if that is unwanted, gate the poll on the release existing first, or shorten `ATTEMPTS` for the failure branch. Either way the job is currently silent in precisely the states it was written to describe.

### 2. minor — dispatching *Release (Rust)* against a tag ref can now cause a crates publish
`.github/workflows/rust-release.yml:250-308`; docs claim at `rust/docs/publishing.md:134`

The docs (added in this PR) restate that "a `workflow_dispatch` run is always a dry run … the workflow cannot be hand-triggered into publishing by mistake." That remains literally true of the publish workflow. It is no longer true of the system: `workflow_dispatch` accepts a tag as its ref, `attach`'s `if: startsWith(github.ref, 'refs/tags/rs-v')` is satisfied, and if no release exists for that tag the new step creates one — which is what publishes. Inferred from reading the trigger and job conditions; I did not dispatch anything.

This is bounded and mostly benign: it needs write access, an existing `rs-v*` tag, a manifest version matching that tag, and no existing release — i.e. it is exactly the recovery action a maintainer would want. But the sentence in the docs now describes only one workflow, not the property a reader will take from it. I would add a clause: dispatching the release workflow *on a tag* is the one hand-trigger that can lead to a publish, deliberately.

### 3. nit — the release workflow's own trigger and ref guard are not asserted
`python/tests/test_rust_release_tag_path.py:158-183`

`CratesPublishStaysTheOneGateTest` pins the publish workflow's `on:` block and the publish job's `if:` and `environment:`. Nothing pins the equivalent for `rust-release.yml`, which is now the workflow that mints releases. I mutation-tested this: adding `pull_request:` to `rust-release.yml`'s triggers, and deleting `attach`'s `if: startsWith(github.ref, 'refs/tags/rs-v')`, both leave all 27 tests green.

Neither mutation is actually dangerous, and that is worth saying: on a pull request `github.ref` is `refs/pull/N/merge`, so `attach` is skipped; and even without the ref guard, the tag-versus-manifest check inside the create step exits 1 before `gh` is called for any ref name that is not `rs-v<manifest version>`. The safety is carried by the version guard, not by the ref guard — good defence in depth. Still, two more assertions would make that intentional rather than incidental.

### 4. nit — `--at` trusts its caller for the version
`.github/scripts/rust/release-decide.py:324-334`

`--at` pins `next_version`/`tag`/notes to whatever string it is handed, with no cross-check against the manifest version it can already compute (`current`). The only caller does check (`rust-release.yml:272-276`), so this is not a live defect — but the flag on its own will happily render notes headed at a version the workspace is not at. A one-line `if args.at != current: sys.exit(...)` would make the invariant local to the flag.

### 5. nit — the shell test writes into the repository root
`python/tests/test_rust_release_tag_path.py:355-360, 393-408`

`CreateStepShellTest` runs the step with `cwd=REPO_ROOT`, so `decision.json` and `release-notes.md` land in the checkout. `setUp` skips if they already exist and `addCleanup` removes them, which is careful, but it makes the test suite mutate the working tree and makes two concurrent runs collide. Copying the scripts' inputs into the tmpdir, or running with `cwd=tmp` and an explicit `-C` on the git calls, would avoid it. Low value; the current form is honest about running the real thing, which is the point.

### 6. nit — the PR body mis-describes the Python path it is mirroring
PR body, "The gap"; `CHANGELOG.md` entry

"Python worked: its workflow creates its own GitHub release, and that release triggers the PyPI publish." The first half is right; the second is not. `python-release.yml:161-165` publishes to PyPI as a step in the same job, gated on the tag ref (and a repository variable), not on a release event. The conclusion the PR draws is unaffected — Python's tag path is self-sufficient — but a reader will come away thinking the two languages share a shape they do not, which matters because the Rust design's release-event indirection is exactly what makes the `crates-io` environment gate possible.

## What is good

Most of this change is right in ways that are easy to get wrong, and several of them I only believed after testing.

- **The wait is the gate, and it is the right gate.** Creating the release is conditioned on the wait's output rather than added beside it, and the wait outlasts the automation's own create by orders of magnitude. I checked the interaction directly: `rust-auto-release.yml:354-355` pushes the tag and `:365` creates the release in the *next step of the same job*, seconds later, while `attach` is behind three cross-platform release builds. There is no double-release path. I could not construct one.
- **`--at` is necessary, not decorative.** I ran both modes against the real range: `--at 0.4.1 --from rs-v0.4.0 --to rs-v0.4.1` renders `## termproof 0.4.1` with `compare/rs-v0.4.0...rs-v0.4.1`; the same range *without* `--at` decides `0.5.0` and would have produced notes headed at a release nobody cut, linking to a tag that does not exist.
- **Failing on a missing token is the correct trade, and it is enforced where it counts.** A release created with the default token cannot start the publish workflow and no re-run recovers it; a tag with no release is recoverable. The guard runs before any `gh` invocation — I confirmed zero `gh` calls on that branch.
- **The tag-versus-manifest check runs before the release exists.** This is the difference between "no release" and "a release the publish workflow will refuse", and it is the second, load-bearing guard noted in finding 3.
- **The verifier derives its crate set and names nothing.** Verified: `verify-crates` reads `publish-plan.py`, exports `ORDER`/`VERSION`, and the script contains none of the three crate directory names. I ran it against the live registry — it passes for the current plan at `0.4.1` (`order: ["termproof"]`, two crates held back) and fails naming the missing crate at `99.99.99`.
- **`5bba742` is a real fix, and I reproduced its verification.** I cloned the branch `--no-tags --depth 1` and executed the create step's own shell against a stub `gh`: it falls through to the whole-history branch, renders notes, and issues exactly one `gh release create rs-v0.4.1 --title rs-v0.4.1 --notes-file release-notes.md --verify-tag`. Under the pre-`5bba742` `$TAG^{commit}` form this path was a hard `rev-parse` failure.
- **A draft release is handled by accident, and well.** If a draft exists, the wait finds it, nothing is created, no publish fires — and the new verifier is exactly what turns that into a red job instead of silence.
- **The tests are not vacuous.** I mutation-tested independently of the PR's claims and reproduced all four it makes: dropping the wait gate (1 error), swapping `secrets.RELEASE_TOKEN` for `github.token` (1 failure), deleting the `verify-crates` job (4 errors), and making the verifier exit 0 on missing crates (2 failures). Removing the tag-versus-manifest guard also fails (2 failures). Lifting the step's `run:` block out of the YAML and executing it under `bash -e` is the reason the suite catches behaviour and not only text, and it is what caught the `$TAG^{commit}` bug in CI.
- **`shellcheck` is clean on the new script.**

## What I could not verify

- That a release created with `RELEASE_TOKEN` emits `release: published` and starts the publish workflow. This is GitHub behaviour, not repository behaviour, and I will not push a tag to test it. It is, however, already proven in production by the weekly path, which creates its release the same way with the same secret.
- The `needs`/`if` skip semantics in finding 1 — read from the workflow and from documented behaviour, not executed.
- The end-to-end timing of `verify-crates` against a real publish. The script polls 80 × 30s ≈ 40 minutes inside a 50-minute job, which should comfortably cover the publish workflow's own gate, but I have no run to measure against. The script's header already flags that a `crates-io` environment with a required reviewer would land here and says to raise `ATTEMPTS` rather than delete the check — that is the right instinct, recorded in the right place.
- The full Python suite locally (the environment here lacks several optional dependencies). I ran the new module — 27 tests, all pass — and relied on CI for the rest, where all three interpreter matrix jobs are green.

## Outside this PR

- **"Nothing else in this repository uploads to a registry"** (`rust-publish-crates.yml:2`) is false as literally written: `python-release.yml:161-165` uploads to PyPI. The claim is pre-existing and is plainly *meant* as "nothing else uploads to crates.io", which is true, and this PR's test only asserts the narrow form against `rust-release.yml`. Worth a two-word edit sometime, not here.
- **`publish.sh` and `verify-published.sh` each carry their own sparse-index URL helper.** The PR body already calls this out and defers it. I agree with the deferral: factoring it would put a structural change to the live publish path inside a behavioural diff.

